PR #48, opened 2020-10-02
Add more abbreivations

ijemmao:
Includes more abbreviations.


shared/utils/replaceAbbreviations.js
```diff
@@ -10,8 +10,10 @@ const abbreviations = {
     'E': 'English',
     'enc': 'enclitic',
     'esp': 'especially',
-    'ext. stuff': 'inflectional suffix',
-    'int': 'interjection',
+    'ext. suff': 'extensional suffix',
+    'H': 'Hausa',
+    'infl. suff': 'inflectional suffix',
+    'int': 'interjection/interrogative',
```
ijemmao, line 16: Since `int.` represents both interject and interrogative, I'm including them both for the same abbreviation.

Issue #92, opened 2020-10-09
Create Separate Usage Docs


ijemmao:
Currently, all the documentation on how to interact with the API is in the README.

That information should be moved out to its own doc. Maybe a USAGE.md file or create a new page in the Wiki.

Maybe we explore using automatically generating API docs with packages like [apidoc](https://www.npmjs.com/package/apidoc)

oemekaogala:
I can take a stab at this!

ijemmao:
This ticket has been addressed by the linked PR.

PR #118, opened 2020-10-14
Added documentation page for API

oemekaogala:
Using swagger.io, I have added documentation for this API. Now we can document the API from our comments and see changes reflected at `http://localhost:8080/docs/`.  

This addresses issue #92.

This is how the docs look.
Screen Shot 2020-10-12 at 8 04 53 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,14 @@ Navigate to [localhost:8080](http://localhost:8080/) to see the API
 
 ## Usage
 
+### Documentation
+
+#### Development
+After starting the API server with this `yarn dev`, visit `http://localhost:8080/docs`.
+
+#### Production
+You can also view the productions docs by visiting `http://igboapi.com/docs`.
+
 ### MongoDB Data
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -52,6 +52,14 @@ Navigate to [localhost:8080](http://localhost:8080/) to see the API
 
 ## Usage
```
ijemmao, line 53: Is there a way we can transfer all the information that's in this `## Usage` section into the swagger site?

That way the main README.md be a lot smaller.

oemekaogala: To my understanding, swagger just allows you to automate the doc itself, so this is supposed to assist the website as a "docs" tab. So when we create the main igboapi.com website, we can add this to that tab and update the elements to match the CSS for the original template.

ijemmao: I see what you mean, so once we have a dedicated docs site we can just change the styles that swagger automatically uses so that it matches the look of our site.

ijemmao (approved):
Looks good!

PR #133, opened 2020-10-15
Added json file for 1000 common used igbo words

pappyJ:
re


src/dictionaries/parseAndBuild-1000.js
```diff
@@ -0,0 +1,101 @@
+const request = require("request");
```
ijemmao, line 1: There's no need to include this script since we aren't going to be running it in the main repo

src/config.js
```diff
@@ -1,10 +1,13 @@
-const DB_NAME = 'igbo_api';
-const TEST_DB_NAME = 'test_igbo_api';
+const DB_NAME = "igbo_api";
```
ijemmao, line 1: It looks like your text editor changed all the single quotes to double-quotes.

Could you change the double-quotes back to single quotes?

ijemmao (changes requested):
One last note, you should rename your dictionary from `ig-en_1000_common.json` and place it in the `ig-en` directory instead of placing it inside the `ig-en-1000` directory that you created.

But so far so good!

pappyJ:
Ohhh my ... It's pretier

pappyJ:
done.. check again

ijemmao (changes requested):
It looks like you're still accidentally formatting files.

You could try pulling from `master` and running `yarn install` so that you can use the pre-commit lint check so that it can catch these formatting errors before you make another push!

src/server.js
```diff
@@ -7,18 +7,19 @@ import { PORT, MONGO_URI } from './config';
 const app = express();
 
 mongoose.connect(MONGO_URI, {
-    useNewUrlParser: true,
-    useCreateIndex: true,
-    useFindAndModify: false
+  useNewUrlParser: true,
```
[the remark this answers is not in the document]
pappyJ: it is caused by prettier ... i'll disable it

pappyJ:
removed the formatting  , just hope it works this time

ijemmao (approved):

ijemmao:
Looks good ✨

Issue #103, opened 2020-10-10
Fix CORS error


ijemmao:
Currently, developers aren't able to directly use the herokuapp endpoints because CORS is denying the client requests.

Allow all domains to be able to access the public API.

ijemmao:
The error still exists

pappyJ:
On it!!

pappyJ:
PR created!!!

PR #146, opened 2020-10-16
allowing cors for http requests 

pappyJ:
fixes issue #103


src/server.js
```diff
@@ -28,6 +28,10 @@ app.use(cors({
   exposedHeaders: ['Content-Range', 'X-Content-Range'],
 }));
 
+//allowing cors on all routes and all request in  pre-flight phase
+
+app.options('*' , cors());
```
ijemmao, line 33: There's already a `cors` object being used on line `27`. You should use that object to configure the CORS settings so we avoid having duplicate objects.

ijemmao (changes requested):
Just a couple of things:

* Write a mocha test to verify that this works
* Please fill out the description of this PR so others know what you did
* Link to the issue that you are closing in the PR description by using GitHub keywords - https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

pappyJ:
Fixes #103

tests/api-mongo.test.js
```diff
@@ -259,3 +328,23 @@ describe('MongoDB Words', () => {
     });
   });
 });
+
+describe('API Requests For Home Directory "/"', () => {
+  describe('/undefinedRoute', () => {
+    it('should return response status of 404', (done) => {
+      chai.request(server).get('/undefined').end((_, res) => {
```
ijemmao, line 335: You should make a new command in the `commands.js` file and then import that new function into this test file so that we can exclude server request logic.

tests/api-mongo.test.js
```diff
@@ -259,3 +328,23 @@ describe('MongoDB Words', () => {
...
+        expect(res.status).to.equal(404);
+        done();
+      });
+    });
+  });
+
+  it('should return hello world', (done) => {
+    chai.request(server).get('/').end((_, res) => {
```
ijemmao, line 343: For this test and the other test, you're testing CORS to see if an application that's running on a different host has the ability to request data from the API's host which is http://igboapi.com

Instead of using `chai.request(server)` which tests the local API, you want to make network requests to `igboapi.com` to see if you're able to still grab data from the site.

ijemmao (changes requested):
You're moving in the right direction with your tests. Just a couple things:

* You're testing the live [API](http://igboapi.com), not the local API. Your tests should point to the Igbo API site
* Move your server request logic into the `commands.js` file
* You included all my commits in this PR. The only files that should be changed are the test and the `command.js` file that you're editing.

ijemmao:
> Fixes #103

Can you please place this in the PR's description instead of a new comment? You can edit the PR description after you've created a new PR

ijemmao:
@pappyJ I'm still see extra files in this PR

It looks like you’re including my commits into your PR (which is resulting in a lot of files that you didn’t touch to look like you edited them)
You want to avoid including other’s commits into your PR so it’s clear what changes you made, you gotta change your git history so that you’re no longer including my commits

PR #180, opened 2020-10-22
Create /DELETE route for Words - resolves ijemmao/igbo_api#131

### LINKED ISSUE #131: Create /DELETE route for Words
Currently, there are some duplicate words in the database, but there's no way to remove those words.

Create a new /DELETE route that a merger would use to delete a word document from the production database.

The route to implement would be DELETE `api/v1/edit/words/:id` which will allow for a single word document to be deleted.

oemekaogala:
In this patch resolves #131. 

I completed two things:

1. the reorganized the documentation comments to align with the corresponding route. This will make it easier for anyone to understand the relationship between the comment/swagger documentation. While going through this documentation, I also noticed there was a missing GET /examples route. The function was already implemented, so I also added this route.

2. Added a new route `api/v1/edit/words/:id` to delete a word from mongodb. I also included documentation for this new route.

**Note**: I decided to return the word as a response when it is successfully deleted

### DIFF
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -199,3 +199,22 @@ export const putWord = (req, res) => {
       return res.send({ error: 'An error has occurred while updating the word, double check your provided data.' });
     });
 };
+
+export const deleteWord = (req, res) => {
+  const { params: { id } } = req;
+
+  return Word.findByIdAndDelete(id)
+    .then(async (word) => {
+      if (!word) {
+        res.status(400);
+        return res.send({ error: 'Word doesn\'t exist' });
+      }
+
+      res.status(200);
+      return res.send(word);
+    })
+    .catch(() => {
+      res.status(400);
+      return res.send({ error: 'An error has occurred while deleting the word, double check your provided id.' });
+    });
+};
--- a/src/routers/router.js
+++ b/src/routers/router.js
@@ -5,96 +5,105 @@ import { getExamples, getExample } from '../controllers/examples';
 const router = express.Router();
 
 /**
- * @swagger
- * /words:
- *   get:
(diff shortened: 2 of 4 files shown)


src/controllers/words.js
```diff
@@ -199,3 +199,22 @@ export const putWord = (req, res) => {
       return res.send({ error: 'An error has occurred while updating the word, double check your provided data.' });
     });
 };
+
+export const deleteWord = (req, res) => {
+  const { params: { id } } = req;
```
ijemmao, line 204: There's a chance that a `WordSuggestion` and `GenericWord` document might reference to a `Word` document with either the `originalWordId` or even the `merged` key. I think we should also make a pass through those documents and remove references from those objects.

src/routers/router.js
```diff
@@ -5,96 +5,105 @@ import { getExamples, getExample } from '../controllers/examples';
 const router = express.Router();
```
ijemmao, line 5: It looks like you didn't create an express route for deleting words yet. Inside `editorRouter.js` could you create the delete route like `editorRouter.delete('/words/:id')`?

ijemmao (changes requested):
This is shaping up! Just a couple things:

* Remove all references to the recently deleted `Word` document inside `WordSuggestion` and `GenericWord` documents
* Create a new express route in `editorRouter` to actually delete a word
* Add Swagger docs to that new route
* Create tests to make sure that not only is the `Word` document actually deleted but there are no more references to that word

oemekaogala:
Okay thanks for the comments. Will work on this

ijemmao:
Closing this PR since it's been inactive. Reopen this PR once you are ready for a new review.

PR #199, opened 2020-10-27
Remove html directory and it content(src/html/*)

ezesundayeze:
closes https://github.com/ijemmao/igbo_api/issues/186


ijemmao:
@ezesundayeze looking good! Could you use the keywords Closes or Fixes instead of Related to so that once this PR is merged that issue gets closed

ijemmao (approved):

PR #206, opened 2020-11-01
Admin grant access

### LINKED ISSUE #125: Create POST/ Routes for Creating new Editor and Merger Auth tokens
Create routes that can be used to create new editor and merger account auth tokens

Blocked by #124 
Blocked by #122

ijemmao:
This PR sets up the groundwork necessary for @ebubae to complete his Firebase auth work.

@ebubae let me know if there's anything I should include or remove from this PR!

Closes #125


src/controllers/admin.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable */
```
ijemmao, line 1: @ebubae This is the file that's responsible for updating user role's

src/controllers/login.js
```diff
@@ -0,0 +1,10 @@
+/* Logs the user in with email and password */
+const login = (req, res) => {
+  // eslint-disable-next-line
+  const { email, password } = req.body;
+
+  // TODO: Get user auth token from Firebase
```
ijemmao, line 6: Here you'll wanna grab the auth token for the user by using their email and password, then send that token back to the frontend so the client can save that token in their cookies.

src/middleware/authentication.js
```diff
@@ -0,0 +1,27 @@
...
+    const { role = 'admin' } = req.query;
+    req.user = { role }; // TODO: Update the shape of this object to match Firebase's user object
+    return next();
+  }
+
+  if (authHeader) {
+    // eslint-disable-next-line
+    const token = authHeader.split(' ')[1];
+
+    // TODO: Verify the Firebase token is valid
```
ijemmao, line 16: Here we want to verify the Firebase token to make sure that it's valid. We don't check it's user role here since that will get checked in `authorization.js`

If the token is valid, place the entire user object on the `req` object like `req.user` = `token.user`, or whatever makes sense.

src/middleware/authorization.js
```diff
@@ -0,0 +1,16 @@
+/* Determines API permission levels based on user role */
```
ijemmao, line 1: This file doesn't need any changes (as far as I know), as long as the authentication file is able to place the user object on the `req` object, then this file should just work out of the box.

src/routers/editorRouter.js
```diff
@@ -20,14 +20,15 @@ import {
...
+import authorization from '../middleware/authorization';
 
 const editorRouter = express.Router();
 
 /* These routes are used to allow users to suggest new words and examples */
-editorRouter.post('/words', mergeWord);
-editorRouter.put('/words/:id', putWord);
-editorRouter.post('/examples', mergeExample);
-editorRouter.put('/examples/:id', putExample);
+editorRouter.post('/words', authorization(['merger', 'admin']), mergeWord);
```
ijemmao, line 28: Only mergers and admins of the platform can merge words and examples into the database.

src/server.js
```diff
@@ -48,12 +56,18 @@ app.use('/api/v1', router);
...
 // TODO: remove this guard rail when releasing for production
 if (process.env.NODE_ENV !== 'production') {
-  app.use('/api/v1', editorRouter);
+  app.post('/login', login);
+  app.use('/admin', authorization(['admin']), adminRouter);
+}
+
+// TODO: remove this guard rail when releasing for production
+if (process.env.NODE_ENV !== 'production') {
+  app.use('/api/v1', authentication, authorization(['editor', 'merger', 'admin']), editorRouter);
```
ijemmao, line 65: Regular users cannot access any of the routes found in `editorRouter`

ijemmao (commented):
@ebubae Hope these comments help with your work.

ebubae (changes requested):
Thanks @ijemmao this is really helpful! Here were the takeways from our discussion on this:

>  We want to call firebase-admin inside authorization.js inside of igbo_api to get the user token, and then place the user object that comes from the token onto the request body in order to grant certain permissions depending on the user role.

This makes sense thanks for explaining.

I think all that's left to do now is remove the logic for logging in and creating the editors as that will be done via firebase.

src/controllers/login.js
ebubae: For now I think we can remove this and use client side routing on the editor page

src/middleware/authentication.js
```diff
@@ -0,0 +1,27 @@
+/* Validates the user-provided auth token */
+const authentication = (req, res, next) => {
+  const authHeader = req.headers.Authorization;
+
+  /* Overrides user role for local development and testing purposes */
+  if (process.env.NODE_ENV !== 'production') {
+    const { role = 'admin' } = req.query;
+    req.user = { role }; // TODO: Update the shape of this object to match Firebase's user object
```
ebubae, line 8: Here I is where we'll be getting the auth token from the query and verifying the token with `firebase-admin`

src/middleware/authorization.js
```diff
@@ -0,0 +1,16 @@
...
+const authorization = (permittedRoles) => (req, res, next) => {
+  const { user } = req;
+
+  /* If the user's role is found in the array of permitted roles,
+   * the user is granted access
+   */
+  if (user && (permittedRoles.includes(user.role) || user.role === 'admin')) {
```
ebubae, line 8: I'm okay with this but cam we remove the last part and just include `admin` as a necessary role. I think it's just better to be explicit so we don't forget how much power admin has/needs to have.

src/middleware/authentication.js
ijemmao: Yeah, but you wouldn't want to put that logic in this `if` block since this section is only for local development and testing purposes.

You would place that logic in the `if` block starting with `if (authHeader)`

src/middleware/authorization.js
ijemmao: When you say include `admin` as a necessary role, do you mean to place it in the `permittedRoles` array?

src/routers/adminRouter.js
```diff
@@ -0,0 +1,8 @@
+import express from 'express';
+import { getUsers } from '../controllers/users';
+
+const adminRouter = express.Router();
+
+adminRouter.get('/users', getUsers);
```
ijemmao, line 6: I created a GET users route because the admin platform would need to get all the users in the database. This will allow an admin to change the user roles as needed.

src/controllers/users.js
```diff
@@ -0,0 +1,6 @@
+/* Get all users from Firebase */
+export const getUsers = (req, res) => {
+  // TODO: Use Firebase functions to get all user objects
+  // This will allow for selecting individual users and updating their user roles
```
ijemmao, line 4: I'm not sure if it's possible, but ideally, we could request all the user objects in Firebase and send that information back to the admin.

src/middleware/authentication.js
ebubae: ahh okay gotcha that makes sense

src/middleware/authorization.js
ebubae: as in just removing the last clause `|| user.role === 'admin')`

src/controllers/users.js
ebubae: I think this is possible with `firebase-admin` https://firebase.google.com/docs/auth/admin/manage-users#list_all_users

ebubae (approved):
Looks good I think we need to update this branch and fix some tests but otherwise is great. 

Thanks for adding all this (especially the extra middleware)!

Issue #238, opened 2020-11-10
Return Response Once If No Term Is Provided for Json data search


pappyJ:
No return keyword is specified when sending the 'NO TERM PROVIDED'  response therefore after sending it ... Code continues to run ... And once it hits the next `res` statement .. Logs `ERR_HTTP_HEADERS_SENT` error to the console

![Screenshot from 2020-11-10 03-30-38](https://user-images.githubusercontent.com/65920240/98620239-3111f080-2305-11eb-90d1-b725089128af.png)
![Screenshot from 2020-11-10 03-29-47](https://user-images.githubusercontent.com/65920240/98620247-366f3b00-2305-11eb-8b67-abb2ee2c95b8.png)
![Screenshot from 2020-11-10 03-30-16](https://user-images.githubusercontent.com/65920240/98620251-3bcc8580-2305-11eb-904b-9a40a934be9f.png)

ijemmao:
@pappyJ Can you please provide a textual description to make this thread more accessible?

pappyJ:
🙄🙄🙄 guidelines please

ijemmao:
@pappyJ A good rule of thumb is to never just add images since there might be a chance that someone who can't see or just render images in their browser can still be apart of the conversation.

Adding text just makes it a lot clearer what the issue is talking about.

With that being said, I think the solution to this problem is to change the line `res.send(NO_PROVIDED_TERM)` to `return res.send(NO_PROVIDED_TERM)` so it doesn't reach the second `res`.

pappyJ:
Exactly ma ...

PR #245, opened 2020-11-11
fixes issue #238 added return keyword to NO PROVIDED TERM response in getWordData method

pappyJ:
fixes issue #238 as described above


src/controllers/words.js
```diff
@@ -28,7 +28,7 @@ export const getWordData = (req, res) => {
   const searchWord = removePrefix(keyword);
   if (!searchWord) {
     res.status(400);
-    res.send(NO_PROVIDED_TERM);
+    return res.send(NO_PROVIDED_TERM);
```
ijemmao, line 31: Can you write a test to make sure that this change works as expected?

ijemmao (changes requested):
Just need to a test for this!

ijemmao:
Closing this PR since it's been inactive. Feel free to reopen it when you are ready for a new review

PR #249, opened 2020-11-13
Auth Middleware

ebubae:
This PR adds middleware for role-based authentication and authorization using Firebase Authentication.

Thanks @ijemmao for fleshing out these modules previously.


ijemmao (approved):
amazing ✨

ijemmao:
@ebubae It looks like a Firebase error is causing the build to fail 😕 - [Failed build](https://github.com/ijemmao/igbo_api/pull/249/checks?check_run_id=1394383188)
```
Error: The default Firebase app already exists. This means you called initializeApp() more than once without providing an app name as the second argument. In most cases you only need to call initializeApp() once. But if you do want to initialize multiple apps, pass a second argument to initializeApp() to give each app a unique name.
```

src/controllers/users.js
```diff
@@ -1,6 +1,14 @@
 /* Get all users from Firebase */
-export const getUsers = (req, res) => {
-  // TODO: Use Firebase functions to get all user objects
+import * as admin from 'firebase-admin';
+
+admin.initializeApp({
+  credential: admin.credential.applicationDefault(),
+});
```
ijemmao, line 6: I see what's going on! Looks you call `admin.initializeApp` twice in two different files.

You should move this block of code to `server.js` where the rest of the third-party packages get initialized. And then delete the identical section in `authentication.js`
A good spot to place this chunk of code could be right above [this line](https://github.com/ijemmao/igbo_api/blob/master/src/server.js#L19)

ijemmao (changes requested):
You can move the Firebase initialization logic into `server.js`

src/middleware/authentication.js
```diff
@@ -1,21 +1,32 @@
...
-    // TODO: Verify the Firebase token is valid
-    // If the token is valid place the user object on the req object
-    // Example: req.user = token.user
+    try {
+      const decoded = await admin.auth().verifyIdToken(token);
+
+      // If the token is valid place the user object on the req object
+      req.user = decoded.user;
+      const userRecord = await admin.auth().getUser(decoded.uid);
+      const userRole = userRecord.customClaims?.role;
```
ijemmao, line 24: Optional chaining isn't enabled for this project 😔

ijemmao (approved):
Once you remove the optional chain, we should be good to go!

PR #261, opened 2020-11-19
mergedBy Fields and Update Auth Middleware

ijemmao:
* There is a new `mergedBy` key that lives on all `wordSuggestion`, `exampleSuggestion`, and `genericWord` documents
    * This will help keep track of who merged the document into a real word or example document
* Update the auth middlware
* Relies on the `FIREBASE_CONFIG` environment variable to use Firebase Admin SDK
* Provides all Firebase users to admins
* Adds tests

### DIFF
--- a/src/controllers/users.js
+++ b/src/controllers/users.js
@@ -1,11 +1,43 @@
 /* Get all users from Firebase */
 import * as admin from 'firebase-admin';
 
-export const getUsers = async (req, res) => {
-  // Use Firebase functions to get all user objects
-  // This will allow for selecting individual users and updating their user roles
-  const result = await admin.default.auth().listUsers();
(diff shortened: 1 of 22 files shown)


src/controllers/users.js
```diff
@@ -1,11 +1,48 @@
...
+    email: user.email || '',
+    displayName: user.displayName || '',
+    role,
+    lastSignInTime: user.metadata.lastSignInTime,
+    creationTime: user.metadata.creationTime,
+  };
+};
+
+/* Grab all users in the Firebase database */
+export const getUsers = async (_, res) => {
```
ijemmao, line 19: @ebubae Turns out that we need to have these functions in the API instead of the editor platform so react-admin can automatically make its pre-built `GET` requests for all users or just a single user.

PR #276, opened 2020-11-29
Igbo api 205/return nothing with no query

### LINKED ISSUE #205: Passing Empty Query String Returns Actual Data In Front End
While this might be good for d Api to return all d words in the DB . I don't Think it is  a good feature for d front end . i headed over [our inspiration website](https://jisho.org/) to confirm and it handles this action very differently
Screenshot from 2020-10-30 01-46-38

ijemmao:
## Background
When a user submits the form with an empty keyword board on the Igbo API homepage, the response will come back with the first ten words of the database. Although this functionality is helpful for admins and developers, it shouldn't be exposed to everyday users of the platform.

## Solution
Depending on the environment and the user's role, return no words or examples if there's no keyword provided.

This PR also:
* Reduce Mocha timeout times from 60 seconds to 10 seconds, speeding up tests
* Deletes `parseHelpers.js` file since it's no longer used

Closes #205


ijemmao:
@pappyJ

PR #295, opened 2020-12-21
Update fallback regex query

ijemmao:
The currently used fallback regex wasn't returning all words and examples. This PR fixes that issue by returning a more loose regex search query.


ijemmao:
No need for this change!

PR #321, opened 2021-02-09
feat: Create Igbo Part of Speech Tagset Constant

### LINKED ISSUE #316: Create Igbo Part of Speech Tagset Constant
## Background
In order for the API to capture the nuance of the Igbo language, it needs to have a part of speech annotating system that provides that nuance. We want to use this API to build treebanks for other projects to use

**Treebank** - A database of sentences that are annotated with syntactic information, often in the form of a tree.
## Feature
Use this [Igbo Part of Speech Tagset](https://www.sketchengine.eu/igbo-part-of-speech-tagset/) annotation system and create a new file called `WordClass` that exports a new object with the following structure

```
export default {
    CJN: {
         label: 'Conjuction',
         description: 'Conjunction'
    }
}
```
`label` should be a simplified label that most people would understand (in this case it's just Conjunction)
`description` should be the description found in the webpage's table

Ebugo:
fixed #316

### DIFF
--- a/src/shared/constants/wordClass.js
+++ b/src/shared/constants/wordClass.js
@@ -1,62 +1,62 @@
 export default {
   ADJ: {
+    value: 'ADJ',
     label: 'Adjective',
-    description: 'Adjective',
   },
   ADV: {
+    value: 'ADV',
     label: 'Adverb',
-    description: 'Adverb',
   },
   CD: {
+    value: 'CD',
     label: 'Number',
-    description: 'Number',
   },
   CJN: {
+    value: 'CJN',
     label: 'Conjunction',
-    description: 'Conjunction',
   },
   DEM: {
+    value: 'DEM',
     label: 'Demonstration',
-    description: 'Demonstration',
   },
   FW: {
+    value: 'FW',
     label: 'Foreign word',
-    description: 'Foreign word',
   },
   INTJ: {
+    value: 'INTJ',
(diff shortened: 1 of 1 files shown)


src/shared/constants/wordClass.js
```diff
@@ -0,0 +1,62 @@
+export default {
+  ADJ: {
+    label: 'Adjective',
+    description: 'Adjective',
```
ijemmao, line 4: Now that I think about this structure, I think it would make more sense to have a `value` field that holds the key value.

For example: 
```
ADJ: {
    value: 'ADJ',
    label: 'Adjective',
}
```

That way when someone indexes into this object, they would be able to get the tagset value along with the label. On top of that, it seems like all the `label` and `description` values are the same so the `description` key will no longer be necessary.

ijemmao (commented):
Just left one comment

src/shared/constants/wordClass.js
Ebugo: Ok, great... I'll fix asap

Ebugo:
> Just left one comment

Seen

src/shared/constants/wordClass.js
Ebugo: Fixed

ijemmao (approved):
Looks good!

Ebugo:
Thanks 😇

PR #361, opened 2021-03-31
Fix incorrect redirect

### LINKED ISSUE #359: Incorrect Link Redirect
**Describe the bug**
The 'dictionary app' link doesn't go to [nkowaokwu.com](https://nkowaokwu.com) instead it goes to localhost:8000.

**To Reproduce**
1. Go to [igboapi.com](https://igboapi.com]
2. Scroll down to the 'Try it Out' section
3. Click on the 'dictionary app' link under the submit button

**Expected behavior**
This link should navigate to `https://nkowaokwu.com`

**Screenshots**

https://user-images.githubusercontent.com/16169291/113162222-f34e2f80-91f3-11eb-9219-b939da639da5.mov

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

gauravsingh5:
Fixes #359

### DIFF
--- a/src/pages/components/Demo/Demo.js
+++ b/src/pages/components/Demo/Demo.js
@@ -19,13 +19,13 @@ const Demo = ({ searchWord, words }) => {
       setProductionUrl(window.origin);
       setInitialQueries(queryString.parse(window.location.search));
       setIsLoading(false);
+      setRedirectUrl(window.location.hostname === 'localhost'
+        ? 'http://localhost:8000'
+        : DICTIONARY_APP_URL);
       if (keyword) {
         window.location.hash = 'try-it-out';
       }
     }
-    setRedirectUrl(window.location.hostname === 'localhost'
-      ? 'http://localhost:8000'
-      : DICTIONARY_APP_URL);
   }, []);
 
   const constructQueryString = () => {
--- a/src/pages/components/Demo/Demo.js
+++ b/src/pages/components/Demo/Demo.js
@@ -4,7 +4,7 @@ import { omit } from 'lodash';
 import queryString from 'query-string';
 import JSONPretty from 'react-json-pretty';
 import { Input, Checkbox } from 'antd';
-import { API_ROUTE, DICTIONARY_APP_URL } from '../../../siteConstants';
+import { API_ROUTE } from '../../../siteConstants';
 
 const Demo = ({ searchWord, words }) => {
   const [isLoading, setIsLoading] = useState(true);
@@ -14,6 +14,7 @@ const Demo = ({ searchWord, words }) => {
   const [productionUrl, setProductionUrl] = useState('');
   const [redirectUrl, setRedirectUrl] = useState('');
   const responseBody = JSON.stringify(words, null, 4);
+  const DICTIONARY_APP_URL = 'https://nkowaokwu.com/';
   useEffect(() => {
     if (typeof window !== 'undefined') {
       setProductionUrl(window.origin);
--- a/src/pages/components/Demo/Demo.js
+++ b/src/pages/components/Demo/Demo.js
@@ -96,7 +96,11 @@ const Demo = ({ searchWord, words }) => {
             </button>
             <p className="text-l text-center text-gray-700 self-center mb-24">
               {'Want to see how this data is getting used? Take a look at the '}
-              <a className="link" href={DICTIONARY_APP_URL}>
+              <a
+                className="link"
+                href={window.location.hostname === 'localhost'
+                  ? 'http://localhost:8000' : DICTIONARY_APP_URL}
+              >
                 dictionary app
               </a>
             </p>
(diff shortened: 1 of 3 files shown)


src/pages/components/Demo/Demo.js
```diff
@@ -96,7 +96,11 @@ const Demo = ({ searchWord, words }) => {
             </button>
             <p className="text-l text-center text-gray-700 self-center mb-24">
               {'Want to see how this data is getting used? Take a look at the '}
-              <a className="link" href={DICTIONARY_APP_URL}>
+              <a
+                className="link"
+                href={window.location.hostname === 'localhost'
```
ijemmao, line 101: I would create a new variable at the top level of the component and then pass into the `href` prop to keep this line clean.

ijemmao (changes requested):
glad to see this happen!

src/pages/components/Demo/Demo.js
```diff
@@ -23,6 +23,9 @@ const Demo = ({ searchWord, words }) => {
         window.location.hash = 'try-it-out';
       }
     }
+    setRedirectUrl(window.location.hostname === 'localhost'
```
ijemmao, line 26: This is might throw an error on the homepage if `window` is undefined (take a look at line 18). `window` might be undefined since this component is rendered on the server side where the `window` object doesn't exist.

If you want to reference `window`, meaning you're in the client, then you want to do so in the `if (typeof window !== 'undefined')` block. But we want to determine the correct link when in the server, where we don't have access to `window`. We would need to do that using environment variables.

So the ideal solution would be, determine the correct link for both the server side and the client side. Take a look at the [NextJS Environment Variable docs](https://nextjs.org/docs/basic-features/environment-variables) for some help

src/siteConstants.js
```diff
@@ -1,7 +1,5 @@
 /* This file includes the constants necessary to power the API homepage */
 export const PORT = process.env.PORT || 8080;
 export const API_ROUTE = process.env.HEROKU ? 'https://igboapi.com' : `http://localhost:${PORT}`;
-export const DICTIONARY_APP_URL = process.env.HEROKU
-  ? 'https://nkowaokwu.com'
-  : 'http://localhost:8000'; // A local instance of the dictionary app must be running
+export const DICTIONARY_APP_URL = 'https://nkowaokwu.com';
```
ijemmao, line 4: We probably want to switch this back to what it was before since it handles the server side part. But inside `Demo.js`, you'll want to move your `setRedirectUrl` into the `if (typeof window !== 'undefined')` block so that you can change the link again on the client side

ijemmao (changes requested):
Unfortunately, you're solution doesn't handle both the client and server side, but leave your questions below and I can help you out ~

gauravsingh5:
What is `process.env.HEROKU` set to? If it is set to anything it'll return true and if it's not set to anything it'll return false. The `API_ROUTE` seems to work but`DICTIONARY_APP_URL` doesn't work.

ijemmao:
> What is `process.env.HEROKU` set to? If it is set to anything it'll return true and if it's not set to anything it'll return false. The `API_ROUTE` seems to work but`DICTIONARY_APP_URL` doesn't work. 

`process.env.HEROKU` is set to true when it's deployed and in production at https://igboapi.com. It's set to false or undefined in every other environment.

gauravsingh5:
Then it should work I guess?

gauravsingh5:
It needs to be set to nothing in any other environment. If the .env file contains `HEROKU=false`, `process.env.HEROKU` will return true because it's set to something.

ijemmao:
> It needs to be set to nothing in any other environment. If the .env file contains `HEROKU=false`, `process.env.HEROKU` will return true because it's set to something.

You're definitely correct, but I that's what I meant, i misspoke. The value of `process.env.HEROKU` will be falsey since it's not set in the development environment.

gauravsingh5:
So we don't need the changes in Demo.js?

ijemmao (commented):
> So we don't need the changes in Demo.js?

We do need changes to `Demo.js`. You change is directly importing `DICTIONARY_APP_URL` that's from `siteConstants.js`. We don't want to refer to any of the variables found in `siteConstant.js` in `Demo.js` since it's always going to give us `http://localhost:8000` when we're in the browser.

I'm recommending creating a new constant with the same name of `DICTIONARY_APP_URL` with the value of `https://nkowaokwu.com` that will live inside `Demo.js`. That is the `DICTIONARY_APP_URL` we want to reference instead of the one that comes from `siteConstants.js`

gauravsingh5:
Should we also remove `DICTIONARY_APP_URL` from `siteConstants.js` because it isn't being used anywhere else other than Demo.js?

ijemmao:
> Should we also remove `DICTIONARY_APP_URL` from `siteConstants.js` because it isn't being used anywhere else other than Demo.js?

Yeah, let's remove it

ijemmao (approved):

ijemmao:
Amazing job! @gauravsingh5

Issue #362, opened 2021-04-07
Search with English using Quotes


ijemmao:
**Is your feature request related to a problem? Please describe.**
When a request for a word is made where the `keyword` is wrapped in quotes, for example, `http://localhost:8080/api/v1/words?keyword="word"`, the API should interpret the term as English and search for related terms.

[The `words` controller](https://github.com/ijemmao/igbo_api/blob/master/src/controllers/words.js#L67-L78) is responsible for determining whether or not to treat the keyword as if it's Igbo or English. The API prioritizes searching with Igbo, so when quotes are involved the API should switch to prioritizing English.

**Describe the solution you'd like**
1. Update the API so that when a keyword is wrapped in quotes, the API will only treat the word as an English term
2. Create tests to make sure that the API handles quoted keywords

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

ijemmao:
This issue has been addressed in #367

PR #367, opened 2021-04-19
feat: add search English by using quotes

Effiti:
solves Issue #362

### DIFF
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -46,9 +46,7 @@ export const searchWordUsingEnglish = async ({ query, searchWord, ...rest }) => 
 /* Gets words from MongoDB */
 export const getWords = async (req, res, next) => {
   try {
-    const hasQuotes = req.query.keyword
-      ? (req.query.keyword.match(/["'].*["']/) !== null)
-      : false;
+    const hasQuotes = req.query.keyword && (req.query.keyword.match(/["'].*["']/) !== null);
     if (hasQuotes) {
       req.query.keyword = req.query.keyword.replaceAll(/["']/g, '');
     }
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -67,7 +67,13 @@ export const getWords = async (req, res, next) => {
     let words;
     let query;
     if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) {
-      const regexKeywordWithoutQuotes = new RegExp((regexKeyword + []).replaceAll(/(["'])|(^\/)|(\/i$)/g, ''), 'i');
+      const regexKeywordWithoutQuotes = new RegExp(
+        (regexKeyword + [])
+          .replaceAll(/["']/g, '')
+          .replaceAll(/^\//g, '')
+          .replaceAll(/\/i$/g, ''),
+        'i',
+      );
       query = searchEnglishRegexQuery(regexKeywordWithoutQuotes);
       words = await searchWordUsingEnglish({ query, ...searchQueries });
     } else {
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -67,10 +67,7 @@ export const getWords = async (req, res, next) => {
     let words;
     let query;
     if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) {
-      const searchWordWithoutQuotes = searchWord
-        .split('')
-        .filter((c) => c !== '"' && c !== "'")
-        .join('');
+      const searchWordWithoutQuotes = searchWord.replaceAll(/[',"]/, '');
       query = searchEnglishRegexQuery(searchWordWithoutQuotes);
       words = await searchWordUsingEnglish({ query, ...searchQueries });
     } else {
@@ -118,12 +115,13 @@ export const getWord = async (req, res, next) => {
       limit: 1,
       dialects,
       examples,
-    }).then(async ([word]) => {
-      if (!word) {
-        throw new Error('No word exists with the provided id.');
-      }
-      return word;
-    });
+    })
+      .then(async ([word]) => {
+        if (!word) {
+          throw new Error('No word exists with the provided id.');
+        }
+        return word;
+      });
     return res.send(updatedWord);
   } catch (err) {
     return next(err);
@@ -142,18 +140,17 @@ export const createWord = async (data) => {
     dialects,
   } = data;
 
-  const emptyDialects = Object.keys(Dialects).reduce(
-    (dialectsObject, key) => ({
-      ...dialectsObject,
-      [key]: {
-        word: '',
-        variations: [],
-        accented: '',
-        dialect: key,
-        pronunciation: '',
-      },
-    }),
-    {},
+  const emptyDialects = Object.keys(Dialects).reduce((dialectsObject, key) => ({
+    ...dialectsObject,
+    [key]: {
+      word: '',
+      variations: [],
+      accented: '',
+      dialect: key,
+      pronunciation: '',
+    },
+  }),
+  {},
   );
 
   const wordData = {
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -80,7 +80,7 @@ export const getWords = async (req, res, next) => {
       : strictSearchIgboQuery(
         searchWord
           .split('')
-          .filter((c) => c !== '"' && c !== "'" && (searchWord.matches(/'.*'/) || searchWord.matches(/'.*'/)))
+          .filter((c) => c !== '"' && c !== "'" && (searchWord.match(/'.*'/) || searchWord.match(/'.*'/)))
           .join(''),
       );
     const words = await getWordSearchFunction(searchWord)({ query, ...searchQueries });
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -48,7 +48,7 @@ export const getWords = async (req, res, next) => {
   try {
     const hasQuotes = req.query.keyword && (req.query.keyword.match(/["'].*["']/) !== null);
     if (hasQuotes) {
-      req.query.keyword = req.query.keyword.replaceAll(/["']/g, '');
+      req.query.keyword = req.query.keyword.replace(/["']/g, '');
     }
     const {
       searchWord,
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -46,7 +46,9 @@ export const searchWordUsingEnglish = async ({ query, searchWord, ...rest }) => 
 /* Gets words from MongoDB */
 export const getWords = async (req, res, next) => {
   try {
-    const hasQuotes = req.query.keyword.match(/["'].*["']/) !== null;
+    const hasQuotes = req.query.keyword
+      ? (req.query.keyword.match(/["'].*["']/) !== null)
+      : false;
     if (hasQuotes) {
       req.query.keyword = req.query.keyword.replaceAll(/["']/g, '');
     }
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -67,7 +67,7 @@ export const getWords = async (req, res, next) => {
     let words;
     let query;
     if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) {
-      const searchWordWithoutQuotes = searchWord.replaceAll(/[',"]/, '');
+      const searchWordWithoutQuotes = searchWord.replaceAll(/[",']/g, '');
       query = searchEnglishRegexQuery(searchWordWithoutQuotes);
       words = await searchWordUsingEnglish({ query, ...searchQueries });
     } else {
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -46,9 +46,9 @@ export const searchWordUsingEnglish = async ({ query, searchWord, ...rest }) => 
 /* Gets words from MongoDB */
 export const getWords = async (req, res, next) => {
   try {
-    const hasQuotes = req.keyword.match(/^["'].*["']$/) !== null;
+    const hasQuotes = req.query.keyword.match(/["'].*["']/) !== null;
     if (hasQuotes) {
-      req.keyword = req.keyword.replaceAll(/["']/g, '');
+      req.query.keyword = req.query.keyword.replaceAll(/["']/g, '');
     }
     const {
       searchWord,
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -44,7 +44,7 @@ export const searchWordUsingEnglish = async ({ query, searchWord, ...rest }) => 
 };
 
 function getWordSearchFunction(searchWord) {
-  if (searchWord.match(/".*"/)) { return searchWordUsingEnglish; }
+  if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) { return searchWordUsingEnglish; }
   return searchWordUsingIgbo;
 }
 
--- a/src/controllers/words.js
+++ b/src/controllers/words.js
@@ -43,6 +43,11 @@ export const searchWordUsingEnglish = async ({ query, searchWord, ...rest }) => 
   return sortDocsBy(searchWord, words, 'definitions[0]');
 };
 
+function getWordSearchFunction(searchWord) {
+  if (searchWord.match(/".*"/)) { return searchWordUsingEnglish; }
+  return searchWordUsingIgbo;
+}
+
 /* Gets words from MongoDB */
 export const getWords = async (req, res, next) => {
   try {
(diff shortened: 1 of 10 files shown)


ijemmao (changes requested):
This looks like it might work! But let's verify it.

Add a test to the `api-mongo.test.js` test file that tests searching for an English word using quotes.

Effiti:
Oops... seems like its searching for the keyword, including the quotes... when searching, it should remove the quotes..... will fix ☺

src/controllers/words.js
```diff
@@ -64,8 +69,21 @@ export const getWords = async (req, res, next) => {
...
       examples,
     };
-    let query = !strict ? searchIgboTextSearch(searchWord, isUsingMainKey) : strictSearchIgboQuery(searchWord);
-    const words = await searchWordUsingIgbo({ query, ...searchQueries });
+    let query = !strict
+      ? searchIgboTextSearch(
+        searchWord
+          .split('')
+          .filter((c) => c !== '"' && c !== "'" && !(searchWord.match(/'.*'/) || searchWord.match(/".*"/)))
+          .join(''),
```
ijemmao, line 77: These lines that are returning a new value that's derived from `searchWord` can be moved out to another function to reduce repetitiveness.

src/controllers/words.js
```diff
@@ -64,8 +69,21 @@ export const getWords = async (req, res, next) => {
       dialects,
       examples,
     };
-    let query = !strict ? searchIgboTextSearch(searchWord, isUsingMainKey) : strictSearchIgboQuery(searchWord);
-    const words = await searchWordUsingIgbo({ query, ...searchQueries });
+    let query = !strict
+      ? searchIgboTextSearch(
+        searchWord
+          .split('')
+          .filter((c) => c !== '"' && c !== "'" && !(searchWord.match(/'.*'/) || searchWord.match(/".*"/)))
```
ijemmao, line 76: The logic that we want to implement with quotes is to prioritize all database queries to be in English if quotes are present in the keyword.
What's happening here is that you are still prioritizing searching with Igbo whether or not quotes are present.

You should:
1. Check to see if the `searchWord` is wrapped in quotes
2. If the `searchWord` is wrapped in quotes, then you want to call `searchEnglishRegexQuery(...)` first, near the top of this function. This means that `words` should should be moved to the top of the function.
3. If there the `searchWord` has no quotes, then you want to go about the original logic of checking the `strict` query ... (but you have to remove the quotes)
4. If `words` turns out to be an empty array, then we fallback to then search with the keyword as English

src/controllers/words.js
```diff
@@ -64,8 +69,21 @@ export const getWords = async (req, res, next) => {
...
+          .join(''),
+        isUsingMainKey,
+      )
+      : strictSearchIgboQuery(
+        searchWord
+          .split('')
+          .filter((c) => c !== '"' && c !== "'" && !(searchWord.match(/'.*'/) || searchWord.match(/".*"/)))
+          .join(''),
```
ijemmao, line 84: This might be more complicated than it needs to be! From the looks of it, you want to remove the quotes from the `searchWord`
I don't think we need these function calls if we are searching with Igbo since Igbo queries should not have quotes in the first place, because in order to get to this part of the function the keyword must not have any quotes.

tests/api-mongo.test.js
```diff
@@ -466,6 +466,30 @@ describe('MongoDB Words', () => {
         });
     });
 
+    it('should return a sorted list of igbo terms when using english by using single qoutes', (done) => {
```
ijemmao, line 469: I would create another test that wraps an Igbo word in quotes and expect to get no results

src/controllers/words.js
```diff
@@ -43,6 +43,11 @@ export const searchWordUsingEnglish = async ({ query, searchWord, ...rest }) =>
   return sortDocsBy(searchWord, words, 'definitions[0]');
 };
 
+function getWordSearchFunction(searchWord) {
```
ijemmao, line 46: We want to create functions like so:
```
const getWordSearchFunction = (searchWord) => {}
```
instead of using the `function` keyword.

src/controllers/words.js
```diff
@@ -43,6 +43,11 @@ export const searchWordUsingEnglish = async ({ query, searchWord, ...rest }) =>
...
+  if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) { return searchWordUsingEnglish; }
```
ijemmao, line 47: Conditional blocks shouldn't be on the same line as the conditional statement. It should look something like this:
```
if (conditionalStatement) {
    return conditionalLogic;
}
```
instead of:
```
if (conditionalStatement) { return conditionalLogic; } 
```

If you want to achieve a single line function, I recommend using ternary operators:

```
const getWordSearchFunction = (searchWord) => (
    conditionalStatement ? searchWordUsingEnglish : searchWordUsingIgbo
)
```

ijemmao (changes requested):
* I noticed that you format your code a little differently, which is fine! But to keep the codebase consistent, you'll need to make a couple small formatting changes
* I explained the logic behind using quotes a little more to give more context. We want to completely prioritize searching in English if our keyword is wrapped in quotes. If there are no quotes, then we go back to the original logic

Effiti:
should it fall back to using igbo when the `searchWord` is wrapped in quotes but it couldnt find any words ? Because then tests that test for no words being found when wrapping igbo terms in quotes would not work.

ijemmao:
> should it fall back to using igbo when the `searchWord` is wrapped in quotes but it couldnt find any words ? Because then tests that test for no words being found when wrapping igbo terms in quotes would not work.

Nope, it shouldn't fallback to searching with Igbo if there were no quotes used. I'm not sure if I'm seeing why the tests would fail when you wrap Igbo words in quotes.

I think it would still be possible to get no results when an Igbo word is wrapped in quotes since we should be using the "search with english" function which doesn't look at head words in the database, rather it should only be looking at the definitions for each word to find matching documents.

src/controllers/words.js
```diff
@@ -64,18 +64,36 @@ export const getWords = async (req, res, next) => {
...
-        res,
-        docs: englishWords,
-        model: Word,
-        query,
-        ...rest,
-      });
+    let words;
+    let query;
+    if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) {
+      const searchWordWithoutQuotes = searchWord
```
ijemmao, line 70: This could be rewritten as:
```
const searchWordWithoutQuotes = searchWord.replaceAll(/['"]/, '')
```
to remove the quotes instead of iterating over each character

src/controllers/words.js
```diff
@@ -100,13 +118,12 @@ export const getWord = async (req, res, next) => {
...
       dialects,
       examples,
-    })
-      .then(async ([word]) => {
-        if (!word) {
-          throw new Error('No word exists with the provided id.');
-        }
-        return word;
-      });
+    }).then(async ([word]) => {
```
ijemmao, line 121: Revert this change since it's unnecessary

src/controllers/words.js
```diff
@@ -125,16 +142,19 @@ export const createWord = async (data) => {
...
-    ...dialectsObject,
-    [key]: {
-      word: '',
-      variations: [],
-      accented: '',
-      dialect: key,
-      pronunciation: '',
-    },
-  }), {});
+  const emptyDialects = Object.keys(Dialects).reduce(
```
ijemmao, line 145: Revert this change since it's unnecessary

tests/api-mongo.test.js
```diff
@@ -466,6 +466,30 @@ describe('MongoDB Words', () => {
         });
     });
 
+    it('should return a list of igbo terms when using english by using single qoutes', (done) => {
```
ijemmao, line 469: Last thing, make sure to add another test that wraps an Igbo word in quotes

ijemmao (commented):
Yeah, this is a lot closer! There are just a few blocks of code that didn't need to be changed, so you should go ahead and revert those changes. And don't forget about that last test 👍🏾

tests/api-mongo.test.js
```diff
@@ -466,6 +466,30 @@ describe('MongoDB Words', () => {
...
+        });
+    });
+
+    it('should also return a list of igbo terms when using english by using double qoutes', (done) => {
+      const keyword = '"water"';
+      getWords({ keyword })
+        .end((_, res) => {
+          expect(res.status).to.be.equal(200);
+          expect(res.body).to.be.an('array');
+          expect(res.body).to.have.lengthOf.at.least(5);
```
ijemmao, line 487: Also, where did the number `5` come from? Why did you go with at least the length of `5`?

src/controllers/words.js
```diff
@@ -67,10 +67,7 @@ export const getWords = async (req, res, next) => {
     let words;
     let query;
     if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) {
-      const searchWordWithoutQuotes = searchWord
-        .split('')
-        .filter((c) => c !== '"' && c !== "'")
-        .join('');
+      const searchWordWithoutQuotes = searchWord.replaceAll(/[',"]/, '');
```
ijemmao: I would double check this regex, but it's also going to target and remove commas

src/controllers/words.js
```diff
@@ -118,12 +115,13 @@ export const getWord = async (req, res, next) => {
       limit: 1,
       dialects,
       examples,
-    }).then(async ([word]) => {
-      if (!word) {
-        throw new Error('No word exists with the provided id.');
-      }
-      return word;
-    });
+    })
```
ijemmao: Again, you'll need to revert this block of code since it's just a formatting change

src/controllers/words.js
```diff
@@ -142,18 +140,17 @@ export const createWord = async (data) => {
...
-      [key]: {
-        word: '',
-        variations: [],
-        accented: '',
-        dialect: key,
-        pronunciation: '',
-      },
-    }),
-    {},
+  const emptyDialects = Object.keys(Dialects).reduce((dialectsObject, key) => ({
```
ijemmao: Revert this formatting change

src/controllers/words.js
```diff
@@ -134,7 +149,9 @@ export const createWord = async (data) => {
       dialect: key,
       pronunciation: '',
     },
-  }), {});
+  }),
```
ijemmao, line 152: Revert this formatting change

tests/api-mongo.test.js
```diff
@@ -466,6 +466,39 @@ describe('MongoDB Words', () => {
...
+          expect(every(res.body, (word) => word.includes('water'))).to.equal(true);
+          done();
+        });
+    });
+
+    it('should not return any words when wrapping an igbo word in quotes', (done) => {
+      const keyword = '"mmili"';
+      getWords({ keyword })
+        .end((_, res) => {
+          expect(res.status).to.be.equal(400);
```
ijemmao, line 497: The response shouldn't be `400` since it's not an error to return an empty list of words. Rather, you should expect a status code of `200` and the `res.body` to have a length of zero

src/controllers/words.js
```diff
@@ -64,18 +64,33 @@ export const getWords = async (req, res, next) => {
...
-        res,
-        docs: englishWords,
-        model: Word,
-        query,
-        ...rest,
-      });
+    let words;
+    let query;
+    if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) {
+      const searchWordWithoutQuotes = searchWord.replaceAll(/[",']/g, '');
```
ijemmao, line 70: I would double check this regex. It looks like it's going to target and remove commas alongside single and double quotes

Effiti:
The error was in calling `searchEnglishRegexQuery` with the `searchWord`, not the` regexKeyword`. ( of course the quotes have to be removed from that)

src/controllers/words.js
```diff
@@ -64,18 +64,39 @@ export const getWords = async (req, res, next) => {
...
-        res,
-        docs: englishWords,
-        model: Word,
-        query,
-        ...rest,
-      });
+    let words;
+    let query;
+    if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) {
+      const regexKeywordWithoutQuotes = new RegExp(
```
ijemmao, line 70: This could be simplified to:

```js
const regexKeywordWithoutQuotes = new RegExp(regexKeyword.toString().replaceAll(/["']/, ''), 'i');
```

This will convert `regexKeyword` into a string, which will allow us to call the `replaceAll` function which will target all the double and single quotes and replace them with empty strings. We then attach the `i` flag for case insensititivity.

src/controllers/words.js
```diff
@@ -64,18 +64,39 @@ export const getWords = async (req, res, next) => {
...
-      return packageResponse({
-        res,
-        docs: englishWords,
-        model: Word,
-        query,
-        ...rest,
-      });
+    let words;
+    let query;
+    if (searchWord.match(/".*"/) || searchWord.match(/'.*'/)) {
```
ijemmao, line 69: So it looks like what you want to do instead, is to remove the quotes from `searchWord` before passing it into `handleQueries` so that the returned `regexKeyword` doesn't include quotes. This will allow you to omit the lines 70 to 76 that you've written since you don't need to convert the regex into a string and back into a regex.

So these are the steps you want to take:
1. Before calling `handleQueries` on line 59, check to see if `searchWord` has either single or double quotes. If it does, keep track of that state by storing the boolean value in a variable (maybe called `hasQuotes`), we will need to know about this later.
2. Remove the quotes from `searchWord`, specifially from `req.searchWord`,  and then call `handleQueries` by passing in `req` that has the newly updated `searchWord` that doesn't have any quotes.
3. This will return `regexKeyword` which is a regular expression that doesn't include any quotes, so we don't have to remove those quotes from the regex later.
4. After we get `regexKeyword` we want to know whether or not our original `searchWord` had quotes or not. If it did, then we want to call `query = searchEnglishRegexQuery()` and return whatever result we get. If it didn't have quotes, then we want search with Igbo first.

src/controllers/words.js
```diff
@@ -46,6 +46,12 @@ export const searchWordUsingEnglish = async ({ query, searchWord, ...rest }) =>
 /* Gets words from MongoDB */
 export const getWords = async (req, res, next) => {
   try {
+    const hasQuotes = req.query.keyword
```
ijemmao, line 49: nit: this can simplified to:
```js
const hasQuotes = req.query.keyword && req.query.keyword.match(/["'].*["']/) !== null);
```
if `req.query.keyword` doesn't exist, this conditional will short circuit and return a falsely value.

src/controllers/words.js
```diff
@@ -64,18 +70,32 @@ export const getWords = async (req, res, next) => {
       dialects,
       examples,
     };
-    let query = !strict ? searchIgboTextSearch(searchWord, isUsingMainKey) : strictSearchIgboQuery(searchWord);
-    const words = await searchWordUsingIgbo({ query, ...searchQueries });
-    if (!words.length) {
+    let words;
+    let query;
+    if (hasQuotes) {
```
ijemmao, line 75: perfect!

tests/api-mongo.test.js
```diff
@@ -466,6 +466,39 @@ describe('MongoDB Words', () => {
         });
     });
 
+    it('should return a list of igbo terms when using english by using single qoutes', (done) => {
```
ijemmao, line 469: you have a typo, it should be `quotes`, instead of `qoutes`

ijemmao (commented):
This is practically good to go! Final steps:
* Address the nit comment
* Fix the typo in the test name
* Revert the change made to `yarn.lock`

tests/api-mongo.test.js
ijemmao: @Effiti Looks like addressing this typo is the last thing!

Effiti: I am sorry.. english is not my main language.. sorry

ijemmao:
@Effiti I suggest running the tests locally to figure out why your three new tests are failing

Effiti:
> @Effiti I suggest running the tests locally to figure out why your three new tests are failing

I dod, and with node version 15 it worked  fine,  ```yarn mocha``` had all tests passing, but with version 12 they failed. 
Also, the tests pass the "keyword" directly as a property of ```req```  not of ```req.query```

ijemmao:
> > @Effiti I suggest running the tests locally to figure out why your three new tests are failing
> 
> I did, and with node version 15 it worked fine, `yarn mocha` had all tests passing, but with version 12 they failed.

You're three tests are failing on 400 network responses, so I don't think the Node version should be the reason to fail the tests. I can try to run the tests locally later today and see what's causing these issues

tests/api-mongo.test.js
```diff
@@ -466,6 +466,39 @@ describe('MongoDB Words', () => {
         });
     });
 
+    it('should return a list of igbo terms when using english by using single quotes', (done) => {
+      const keyword = '\'water\'';
+      getWords({ keyword })
+        .end((_, res) => {
+          expect(res.status).to.be.equal(200);
```
ijemmao, line 473: This is the error I'm getting when I console `res.body`:

Screen Shot 2021-04-26 at 10 14 23 AM

Turns out that `replaceAll` doesn't work in the current working version of Node 12 and 14 (the function might be available in Node 15)

To get around this, use `replace` instead of `replaceAll` and use a RegExp:

```
req.query.keyword.replace(/["']/g, '');
```

ijemmao (approved):

Issue #627, opened 2023-05-09
create developer information endpoint


davydocsurg:
The developer information endpoint will be responsible for fetching a logged in developer's details

PR #634, opened 2023-05-15
627 create developer information endpoint

davydocsurg:
## Describe your changes
created developer information endpoint

#627 

## How Has This Been Tested?
Tested with Postman

Screenshot 2023-05-15 at 12 05 31 PM


src/controllers/developers.js
```diff
@@ -62,3 +58,23 @@ export const postDeveloper = async (req, res, next) => {
     return next(err);
   }
 };
+
+export const getDeveloper = async (req, res, next) => {
+  const connection = createDbConnection();
+  const Developer = connection.model('Developer', developerSchema);
+  const devId = req.params.id;
```
ijemmao, line 65:
```suggestion
  const { id } = req.params;
```

src/controllers/developers.js
```diff
@@ -62,3 +58,23 @@ export const postDeveloper = async (req, res, next) => {
...
+  try {
+    const developer = await Developer.find({ _id: devId });
```
ijemmao, line 67: we have a custom Mongoose plugin responsible for applying the `id` field to the found `developer` object. So we would want to do the following to remove the default `_id` key with `id`
```suggestion
    let developer = await Developer.find({ _id: devId });
    
    if (developer) {
      throw new Error('Developer doesn\'t exist');
    }

    developer = developer.toJSON();
```

src/routers/router.js
```diff
@@ -29,6 +29,7 @@ router.get('/examples', validateApiKey, attachRedisClient, getExamples);
 router.get('/examples/:id', validateApiKey, validId, attachRedisClient, getExample);
 
 router.post('/developers', createDeveloperLimiter, validateDeveloperBody, postDeveloper);
+router.get('/developers/:id', validId, attachRedisClient, getDeveloper);
```
ijemmao, line 32: make sure to test your changes locally. the middleware `validId` wouldn't work since it only works for MongoDB document Ids. You will probably want to create a new middleware like `authorization` that checks the user's API token (similar to what `validateApiKey` does) but then checks that the provided API key is associated with the requested use with `:id`

davydocsurg:
> this is great! you'll need to write two tests as well:
> 
> * check to see if a developer document is returned with the correct credentials
> * check to see if a 403 status response is returned with the incorrect credentials

Okay, on it

.gitignore
```diff
@@ -149,4 +149,7 @@ functions/node_modules
 functions/src
 functions/swagger.json
 functions/yarn.lock
-yarn.lock
\ No newline at end of file
+yarn.lock
+
+# prettier
+.prettierignore
```
ijemmao, line 155: remove this line since it doesn't relate to your changes - if you want to make this change (which I encourage!) you can open a separate PR

src/controllers/developers.js
```diff
@@ -62,3 +58,29 @@ export const postDeveloper = async (req, res, next) => {
...
+  const Developer = connection.model('Developer', developerSchema);
+  const { id } = req.params;
+  try {
+    const developer = await Developer.find({ _id: id });
+
+    if (!developer.length) {
+      throw new Error("Developer doesn't exist");
+    }
+
+    // developer = developer.toJSON();
```
ijemmao, line 73: why is this line commented?

src/controllers/developers.js
```diff
@@ -62,3 +58,29 @@ export const postDeveloper = async (req, res, next) => {
     return next(err);
   }
 };
+
+export const getDeveloper = async (req, res, next) => {
+  const connection = createDbConnection();
+  const Developer = connection.model('Developer', developerSchema);
+  const { id } = req.params;
+  try {
+    const developer = await Developer.find({ _id: id });
```
ijemmao, line 67:
```suggestion
    const developer = await Developer.findById(id);
```

src/controllers/developers.js
```diff
@@ -62,3 +58,29 @@ export const postDeveloper = async (req, res, next) => {
...
+
+    if (!developer.length) {
+      throw new Error("Developer doesn't exist");
+    }
```
ijemmao, line 71: we only expect one developer to match the provided id, so we can do something like this:
```suggestion
    if (!developer) {
      throw new Error("Developer doesn't exist");
    }
```

src/controllers/developers.js
```diff
@@ -62,3 +58,29 @@ export const postDeveloper = async (req, res, next) => {
...
+
+export const getDeveloper = async (req, res, next) => {
+  const connection = createDbConnection();
+  const Developer = connection.model('Developer', developerSchema);
+  const { id } = req.params;
+  try {
+    const developer = await Developer.find({ _id: id });
+
+    if (!developer.length) {
+      throw new Error("Developer doesn't exist");
```
ijemmao, line 70: nit: please use single quotes
```suggestion
      throw new Error('Developer doesn\'t exist');
```

src/middleware/authorization.js
```diff
@@ -0,0 +1,21 @@
...
+export default async (req, res, next) => {
+  try {
+    // const { apiLimit } = req.query;
+    let apiKey = fetchAPIKey(req);
+
+    if (!apiKey && isDevelopment) {
+      apiKey = FALLBACK_API_KEY;
+    }
+
+    //   check if api key belongs to developer
```
ijemmao, line 13: nit: remove the extra space between `//` and `check`

src/middleware/authorization.js
```diff
@@ -0,0 +1,21 @@
+import { isDevelopment } from '../config';
+import { FALLBACK_API_KEY, fetchAPIKey, findDeveloper } from '../shared/constants/DeveloperUtils';
+
+export default async (req, res, next) => {
+  try {
+    // const { apiLimit } = req.query;
```
ijemmao, line 6: why is this line commented?

src/shared/constants/DeveloperUtils.js
```diff
@@ -0,0 +1,81 @@
+import { compareSync } from 'bcrypt';
```
ijemmao, line 1: this file isn't a constants file - rather this file should live closer to the developers controller file. Yous should create a new `developers` directory that includes `developers.js` and `utils.js` where the utils file includes all these functions

src/shared/constants/DeveloperUtils.js
```diff
@@ -0,0 +1,81 @@
...
+    await handleDeveloperUsage(developer);
+    return next();
+  }
+
+  res.status(401);
+  return res.send({ error: 'Your API key is invalid' });
+};
+
+// eslint-disable-next-line
+export { fetchAPIKey, findDeveloper, fetchRequestQuery, checkDeveloperAPIKey, FALLBACK_API_KEY };
```
ijemmao, line 81: instead of exporting this object at the end of the file, you can prepend the `export` keyword before each variable when they are getting created

src/shared/constants/DeveloperUtils.js
```diff
@@ -0,0 +1,81 @@
...
+// const isSameDate = (first, second) =>
+//   first.getFullYear() === second.getFullYear() &&
+//   first.getMonth() === second.getMonth() &&
+//   first.getDate() === second.getDate();
+
+const isSameDate = (first, second) => {
+  const firstISO = first.toISOString().slice(0, 10); // Extract YYYY-MM-DD from first date
+  const secondISO = second.toISOString().slice(0, 10); // Extract YYYY-MM-DD from second date
+  return firstISO === secondISO;
+};
```
ijemmao, line 24: please revert these change for this PR - if you want to make this change in a future PR with an explanation on why you are making this change to `isSameDate` then you can do so

src/shared/constants/DeveloperUtils.js
ijemmao, line 1: it's not really recommended to move a file and then make changes within that moved file since it's difficult to spot changes you actually made (since the entire file is now green in GitHub) - ideally, if you want to refactor the codebase, you would just move the file.

please only move the file in this PR and then in future PRs you can suggest making changes

ijemmao (changes requested):

.gitignore
davydocsurg: Okay, I did this to avoid some conflicts I got when I saved the file but I can remove it now

src/controllers/developers.js
davydocsurg: It throws this error:
`
{
    "error": "_developer2.toJSON is not a function"
}
`

src/middleware/authorization.js
davydocsurg: fixed in the `fetchAPIKey()`. Removed

src/shared/constants/DeveloperUtils.js
davydocsurg: Okay

src/shared/constants/DeveloperUtils.js
davydocsurg: Done

src/shared/constants/DeveloperUtils.js
davydocsurg: Done

src/shared/constants/DeveloperUtils.js
davydocsurg: Okay

__tests__/shared/commands.js
```diff
@@ -101,3 +101,10 @@ export const searchMockedTerm = (term) => {
   const { wordReg: regexTerm } = createRegExp(term);
   return resultsFromDictionarySearch(regexTerm, term, mockedData);
 };
+
+/* fetch developer details */
+export const getDeveloper = (id, options = {}) => (
+  server
+    .get(`${API_ROUTE}/developers/${id}`)
+    .set('X-API-Key', options.apiKey || FALLBACK_API_KEY)
+);
```
ijemmao, line 110: nit: add an extra line at the end of this file

src/controllers/developers/developers.js
```diff
@@ -62,3 +58,28 @@ export const postDeveloper = async (req, res, next) => {
...
+
+export const getDeveloper = async (req, res, next) => {
+  const connection = createDbConnection();
+  const Developer = connection.model('Developer', developerSchema);
+  const { id } = req.params;
+  try {
+    const developer = await Developer.findById(id);
+
+    if (!developer) {
+      throw new Error("Developer doesn't exist");
```
ijemmao, line 70: nit: please use single quotes instead of double quotes
```suggestion
      throw new Error('Developer doesn\'t exist');
```

src/middleware/authorization.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default async (req, res, next) => {
+  try {
+    let apiKey = fetchAPIKey(req);
+
+    if (!apiKey && isDevelopment) {
+      apiKey = FALLBACK_API_KEY;
+    }
+
+    // check if api key belongs to developer
```
ijemmao, line 12: nit:

```suggestion
    // Check if API key belongs to the requested developer
```

src/middleware/authorization.js
```diff
@@ -0,0 +1,20 @@
...
+    await findDeveloper(apiKey);
+
+    return next();
+  } catch (err) {
+    res.status(400);
+    return res.send({ error: err.message });
```
ijemmao, line 18: nit:

```suggestion
    res
      .status(400)
      .send({ error: err.message });
```

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,72 @@
...
+  return updatedDeveloper.save();
+};
+
+export const fetchAPIKey = (req) => {
+  const apiKey = req.headers['X-API-Key'] || req.headers['x-api-key'];
+  if (!apiKey) {
+    throw new Error("X-API-Key Header doesn't exist");
+  }
+  return apiKey;
+};
```
ijemmao, line 41: again, i don't think we should adding new logic this file - rather you should just be moving already existing utils to this file

the reason for this is because this function `fetchAPIKey` will short circuit the logic where a developer is running the project locally but doesn't have an API Key in their header - it will cause the server to return an error which is unexpected behavior

i would be open to having this change in a separate PR but not in this one, please revert this util since it changes the already existing functionality that shouldn't be changed within this PR

src/middleware/validateApiKey.js
```diff
@@ -64,28 +24,16 @@ export default async (req, res, next) => {
     if (!apiKey && isDevelopment) {
       apiKey = FALLBACK_API_KEY;
     }
-    if (!apiKey) {
-      throw new Error('X-API-Key Header doesn\'t exist');
-    }
```
ijemmao, line 69: these lines should be left alone

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,72 @@
...
+import { isTest } from '../../config';
+
+const PROD_LIMIT = 2500;
+export const FALLBACK_API_KEY = 'fallback_api_key';
+
+const determineLimit = (apiLimit) => (isTest ? apiLimit || PROD_LIMIT : PROD_LIMIT);
+export const fetchRequestQuery = (req) => {
+  const { apiLimit } = req.query;
+  return apiLimit;
+};
```
ijemmao, line 12: this doesn't need to be a util, you could just reference the apiLimit directly in the file that needs it for now

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,72 @@
...
+    if (developer.usage.count >= determineLimit(apiLimit)) {
+      res.status(403);
+      return res.send({ error: 'You have exceeded your limit of requests for the day' });
+    }
+    await handleDeveloperUsage(developer);
+    return next();
+  }
+
+  res.status(401);
+  return res.send({ error: 'Your API key is invalid' });
```
ijemmao, line 71: helper functions shouldn't be responsible for sending a request back to the client - it should just be responsible for performing its base logic and then returning that value to the outer function that called this one. the outer function will then do a `res.send()`

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,72 @@
...
+
+export const checkDeveloperAPIKey = async (req, res, next) => {
+  const { apiLimit } = fetchRequestQuery(req);
+  const apiKey = fetchAPIKey(req);
+  const developer = await findDeveloper(apiKey);
+
+  if (developer) {
+    if (developer.usage.count >= determineLimit(apiLimit)) {
+      res.status(403);
+      return res.send({ error: 'You have exceeded your limit of requests for the day' });
```
ijemmao, line 64: instead of a doing a `send.status().send()` here, you would want to throw an error that would be caught by the outer method that will then handle sending the 403 status code and response body back to the client

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+    await handleDeveloperUsage(developer);
+    return next();
```
ijemmao, line 67: again, you don't want to really reference any express arguments here (that includes `req`, `res, `next`) since you're calling `next()` in this function and in the outer method

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,72 @@
...
+  const Developer = connection.model('Developer', developerSchema);
+  const developer = await Developer.find({ apiKey });
+  if (!developer.length) {
+    throw new Error('Invalid API Key. Check your API Key and try again');
+  }
+  console.timeEnd('Finding developer account');
+  return developer;
+};
+
+export const checkDeveloperAPIKey = async (req, res, next) => {
```
ijemmao, line 56: you don't want to pass in any Express arguments since it makes it more difficult to write unit tests for this function in the future. Instead, you will just want to pass in the important arguments to allow this function to run like `apiLimit` and `apiKey`, for instance

ijemmao (changes requested):
making progress!

src/controllers/developers/utils.js
davydocsurg: Okay

.gitignore
ijemmao: @davydocsurg any update on removing these prettier lines? let's focus on this in another PR

__tests__/developers.test.js
```diff
@@ -110,5 +97,18 @@ describe('Developers', () => {
       const res = await getWord(limitWordId, { apiLimit: 2 }, { apiKey: developerRes.body.apiKey });
       expect(res.status).toEqual(403);
     });
+
+    it('should return developer document with correct credentials', async () => {
+      const developerRes = await createDeveloper(developerData);
+      const developerDetails = await getDeveloper({ apiKey: developerRes.body.apiKey });
+      expect(developerDetails.status).toEqual(200);
+      expect(developerDetails.body).not.toEqual(undefined);
```
ijemmao, line 105: @davydocsurg instead of checking to see if the body is undefined, can you check to see if the body includes the correct data?

__tests__/shared/commands.js
```diff
@@ -14,6 +14,7 @@ import mockedData from '../__mocks__/data.mock.json';
 const server = request(app);
 
 export const createDeveloper = (data) => (
+  // console.warn('createDeveloper is deprecated. Use createDeveloperV2 instead.');
```
ijemmao, line 17: remove this comment

src/middleware/authorization.js
```diff
@@ -0,0 +1,24 @@
...
+      apiKey = FALLBACK_API_KEY;
+    }
+
+    // Check if API key belongs to the requested developer
+    const developer = await findDeveloper(apiKey);
+
+    if (developer.length < 1) {
+      throw new Error('Invalid API Key. Check your API Key and try again', 403);
+    }
+    res.locals.developer = developer;
```
ijemmao, line 18: i don't think we want to start the pattern of adding objects to the `res` Response object, instead you would want to do something like the following that saves the data on the `req` Request object
```suggestion
    req.locals.developer = developer;
```

src/middleware/validateApiKey.js
ijemmao: @davydocsurg bumping this message

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,64 @@
...
+  const Developer = connection.model('Developer', developerSchema);
+  const developer = await Developer.find({ apiKey });
+  if (!developer) {
+    throw new Error('Invalid API Key. Check your API Key and try again');
+  }
+  console.timeEnd('Finding developer account');
+  return developer;
+};
+
+export const checkDeveloperAPIKey = async (apiLimit, apiKey, next) => {
```
ijemmao, line 51: change the name of this function to be more clear about it's functionality - it should be more like `checkAPILimit` or something

ijemmao (changes requested):

__tests__/developers.test.js
davydocsurg: Alright

__tests__/shared/commands.js
davydocsurg: Done

src/controllers/developers/utils.js
davydocsurg: Done

__tests__/developers.test.js
```diff
@@ -110,5 +97,18 @@ describe('Developers', () => {
       const res = await getWord(limitWordId, { apiLimit: 2 }, { apiKey: developerRes.body.apiKey });
       expect(res.status).toEqual(403);
     });
+
+    it('should return developer document with correct credentials', async () => {
+      const developerRes = await createDeveloper(developerData);
+      const developerDetails = await getDeveloper({ apiKey: developerRes.body.apiKey });
+      expect(developerDetails.status).toEqual(200);
+      expect(developerDetails.body.message).toEqual('Developer retrieved!');
```
ijemmao, line 105: instead of checking the `message` field, can you check each field on the returned developer document to see if it's in the right shape with the right data?

src/controllers/developers/developers.js
```diff
@@ -62,3 +58,19 @@ export const postDeveloper = async (req, res, next) => {
     return next(err);
   }
 };
+
+export const getDeveloper = async (req, res, next) => {
+  try {
+    const { developer } = res.locals;
+
+    return res.status(200).send({
+      message: 'Developer retrieved!',
```
ijemmao, line 67: there's no need for this `message` field since we have a response status of 200 and the developer body, you can remove `message`

src/middleware/validateApiKey.js
```diff
@@ -1,58 +1,19 @@
...
+  checkDeveloperAPIKey,
+  findDeveloper,
+} from '../controllers/developers/utils';
 
 export default async (req, res, next) => {
   try {
-    const { apiLimit } = req.query;
-    let apiKey = req.headers['X-API-Key'] || req.headers['x-api-key'];
+    let { apiToken: apiKey, apiLimit } = handleRequest(req);
+    apiLimit = 2500;
```
ijemmao, line 16: why are you setting the apiLimit again? i don't think you need to explore `apiLimit` from `handleRequest` since `PROD_LIMIT` is the constant that determines the limit for API usage

src/middleware/validateApiKey.js
```diff
@@ -1,58 +1,19 @@
...
-  const Developer = connection.model('Developer', developerSchema);
-  const developers = await Developer.find({});
-  const developer = developers.find((dev) => compareSync(apiKey, dev.apiKey));
-  console.timeEnd('Finding developer account');
-  return developer;
-};
+import {
+  handleRequest,
+  FALLBACK_API_KEY,
+  checkDeveloperAPIKey,
```
ijemmao, line 9: does `checkDeveloperAPIKey` exist in `../controllers/developers/utils`?

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,64 @@
...
+  const Developer = connection.model('Developer', developerSchema);
+  const developer = await Developer.find({ apiKey });
+  if (!developer) {
+    throw new Error('Invalid API Key. Check your API Key and try again');
+  }
+  console.timeEnd('Finding developer account');
+  return developer;
+};
+
+export const checkDeveloperAPILimit = async (apiLimit, apiKey, next) => {
```
ijemmao, line 51: again, your utils functions shouldn't be concerned with Express functionality (i.e. `res`, `req`, `next`)

this function should be responsible for throwing errors if necessary, and then the outer function calling `checkDeveloperAPILimit` should catch the error and handle it accordingly - let's not pass in `next` into this function

ijemmao, line 51: you don't need to pass in the `apiLimit` as an argument, since you have access to the actual API limit `PROD_LIMIT`

src/middleware/authorization.js
```diff
@@ -0,0 +1,24 @@
...
+
+    // Check if API key belongs to the requested developer
+    const developer = await findDeveloper(apiKey);
+
+    if (developer.length < 1) {
+      throw new Error('Invalid API Key. Check your API Key and try again', 403);
+    }
+    req.locals.developer = developer;
+    next();
+    return;
```
ijemmao, line 20:
```suggestion
    return next();
```

src/middleware/authorization.js
```diff
@@ -0,0 +1,24 @@
...
+  } catch (err) {
+    res.status(400).send({ error: err.message });
```
ijemmao, line 22:
```suggestion
return res.status(400).send({ error: err.message });
```

ijemmao (changes requested):

__tests__/developers.test.js
davydocsurg: Done

src/middleware/validateApiKey.js
davydocsurg: Fixed

.gitignore
davydocsurg: Done

firebase.json
```diff
@@ -5,16 +5,14 @@
...
-      "region": "us-central1"
-    }],
+    "ignore": ["firebase.json", "**/.*", "**/node_modules/**"],
+    "rewrites": [
+      {
+        "source": "**",
+        "function": "api",
+        "region": "us-central1"
+      }
+    ],
```
ijemmao, line 15: nit: revert these changes since you're not changing the functionality and to reduce the number of diffs

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,64 @@
...
+  if (!developer) {
+    throw new Error('Invalid API Key. Check your API Key and try again');
+  }
+  console.timeEnd('Finding developer account');
+  return developer;
+};
+
+export const checkDeveloperAPILimit = async (apiKey) => {
+  const developer = await findDeveloper(apiKey);
+  console.log(developer);
```
ijemmao, line 52: remove

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,64 @@
...
+
+  if (developer) {
+    console.log(determineLimit(PROD_LIMIT));
```
ijemmao, line 55: remove

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,64 @@
...
+    if (developer.usage.count >= determineLimit(PROD_LIMIT)) {
+      throw new Error('You have exceeded your API limit. Please upgrade your plan.', 403);
```
ijemmao, line 57: according to the [Error class docs page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error) the second argument of this class is supposed to be an object, we would want to do something like this:

```suggestion
      throw new Error('You have exceeded your API limit. Please upgrade your plan.', { cause: 403 });
```

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+    await handleDeveloperUsage(developer);
+    return developer;
+  }
+
+  throw new Error('Invalid API Key. Check your API Key and try again', 401);
```
ijemmao, line 63: same thing here for the second argument in this new Error

src/middleware/authorization.js
```diff
@@ -0,0 +1,24 @@
...
+
+    if (!apiKey && isDevelopment) {
+      apiKey = FALLBACK_API_KEY;
+    }
+
+    // Check if API key belongs to the requested developer
+    const developer = await findDeveloper(apiKey);
+
+    if (developer.length < 1) {
+      throw new Error('Invalid API Key. Check your API Key and try again', 403);
```
ijemmao, line 16: same comment

src/middleware/authorization.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+    console.info('Developer found', developer);
+    req.developer = developer;
+    return next();
+  } catch (err) {
+return res.status(400).send({ error: err.message });
```
ijemmao, line 22: it seems like you're not using the status codes when you catch the errors that include a status code - you would want to do something like this

```suggestion
let status = 400;
if ('cause' in err) {
  status = cause;
}
return res.status(status).send({ error: err.message });
```

ijemmao (changes requested):

src/controllers/developers/utils.js
davydocsurg: Okay

src/middleware/authorization.js
```diff
@@ -0,0 +1,28 @@
...
+      apiKey = FALLBACK_API_KEY;
+    }
+
+    // Check if API key belongs to the requested developer
+    const developer = await findDeveloper(apiKey);
+
+    if (developer.length < 1) {
+      throw new Error('Invalid API Key. Check your API Key and try again', { cause: 403 });
+    }
+    console.info('Developer found', developer);
```
ijemmao, line 18: remove

src/middleware/validateApiKey.js
ijemmao: @davydocsurg bumping this message again - can you respond mentioning that you've addressed this?

__tests__/developers.test.js
```diff
@@ -110,5 +97,30 @@ describe('Developers', () => {
...
+          date: expect.any(String),
+          count: expect.any(Number),
+        }),
+        name: expect.any(String),
+        apiKey: expect.any(String),
+        email: expect.any(String),
+        password: expect.any(String),
+        createdAt: expect.any(String),
+        updatedAt: expect.any(String),
+        id: expect.any(String),
```
ijemmao, line 116: you want to expect the specific data that you provided to create developer - specifically you want to check to make sure the usage.count is 0, name and email is what you provided

ijemmao (changes requested):

PR #636, opened 2023-05-20
614 migrate to typescript

### LINKED ISSUE #614: Create `tsconfig.json` file
We want to convert the Igbo API codebase to start using Typescript.

Create a new `tsconfig.json` file that will establish the base configuration for our TS codebase.

N/A

**Additional context**
N/A

Ebugo:
## Describe your changes
Add base setup for project migration to typescript

fix #614 

## How Has This Been Tested?
n pm test


.eslintrc.js
```diff
@@ -32,4 +32,58 @@ module.exports = {
...
+        '@next/next/no-html-link-for-pages': ['off'],
+        '@next/next/no-img-element': ['off'],
+        'linebreak-style': ['error', os.platform() === 'win32' ? 'windows' : 'unix'],
+        '@typescript-eslint/no-unused-vars': [
+          'warn',
+          {
+            varsIgnorePattern: '^_',
+            argsIgnorePattern: '^_',
+          },
+        ],
```
ijemmao, line 83: why not place all these rules in the rules section above? what's the purpose of the overrides section?

.eslintrc.js
```diff
@@ -32,4 +32,58 @@ module.exports = {
...
     'linebreak-style': ['error', os.platform() === 'win32' ? 'windows' : 'unix'],
   },
+  overrides: [
+    {
+      files: ['**/*.ts', '**/*.tsx'],
+      env: {
+        es2021: true,
+        node: true,
+        jest: true,
+      },
```
ijemmao, line 42: this is duplicated to the `env` section above

.eslintrc.js
```diff
@@ -32,4 +32,58 @@ module.exports = {
...
+      extends: [
+        'next',
+        'airbnb',
+        'airbnb-typescript',
+        'plugin:cypress/recommended',
+        'eslint:recommended',
```
ijemmao, line 48: do we need to add `eslint:recommended` what benefits are we getting that we weren't getting already?

.eslintrc.js
```diff
@@ -32,4 +32,58 @@ module.exports = {
...
+        'plugin:@typescript-eslint/eslint-recommended',
+        'plugin:@typescript-eslint/recommended',
+      ],
+      parser: '@typescript-eslint/parser',
+      parserOptions: {
+        ecmaVersion: 12,
+        sourceType: 'module',
+        project: ['./tsconfig.json'],
+      },
```
ijemmao, line 57: why is this section here in the overrides section? i think you can just updated the existing fields above

@types/console.d.ts
```diff
@@ -0,0 +1,11 @@
+declare global {
+  interface Console {
+    green(message?: any, ...optionalParams: any[]): void;
+    blue(message?: any, ...optionalParams: any[]): void;
+    red(message?: any, ...optionalParams: any[]): void;
+  }
+}
+
+// If this file has no import/export statements (i.e. is a script)
+// convert it into a module by adding an empty export statement.
```
ijemmao, line 10: nit: remove these comments

@types/environment.d.ts
```diff
@@ -0,0 +1,11 @@
+declare global {
+  namespace NodeJS {
+    interface ProcessEnv {
+      readonly NODE_ENV: 'development' | 'production' | 'test' | 'build';
+    }
+  }
+}
+
+// If this file has no import/export statements (i.e. is a script)
+// convert it into a module by adding an empty export statement.
```
ijemmao, line 10: nit: remove this comment

src/server.ts
```diff
@@ -1,11 +1,12 @@
 import { PORT } from './config';
 import app from './app';
+import NodeEnv from './shared/constants/NodeEnv';
```
ijemmao, line 3: we wouldn't want to create a new object - instead we would want to update the global type of `process.env.NODE_ENV` so that when we attempt to do something like `process.env.NODE_ENV === 'build'` we don't get a TS error

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "outDir": "./build",
+    "allowJs": true,
+    "target": "es5",
+    "lib": ["dom", "dom.iterable", "esnext"],
```
ijemmao, line 6: i think we would just need `esnext` right? maybe just `dom` too?

ijemmao (changes requested):
great start!

Ebugo:
> great start!

Thanks!

.eslintrc.js
Ebugo: > why not place all these rules in the rules section above? what's the purpose of the overrides section?

I used that because most of the project is still common js, so there'd be conflicts if I don't use separate rules for the linting. We'd just end up having erros everywhere.

@types/environment.d.ts
Ebugo: Okay, taking those out now

src/server.ts
Ebugo: That'd cause an error just as I mentioned on Slack, extending the interface ends in errors.

@types/environment.d.ts
Ebugo: I've actually taken off this environment type override entirely due to the type conflicts I mentioned.

tsconfig.json
Ebugo: Okay, on it

.eslintrc.js
```diff
@@ -32,4 +32,35 @@ module.exports = {
...
+          {
+            varsIgnorePattern: '^_',
+            argsIgnorePattern: '^_',
+          },
+        ],
+        '@typescript-eslint/no-empty-function': ['warn'],
+        'prefer-const': 'error',
+      },
+    },
+  ],
```
ijemmao, line 65: this `overrides` section doesn't really need to be here - let's prioritize including all these configurations in the original body of this eslintrc file

@types/console.d.ts
ijemmao: @Ebugo

src/server.ts
ijemmao: @Ebugo have you tried this? https://stackoverflow.com/a/53981706

ijemmao (changes requested):

Ebugo:
That was the very first step I took.. view the error

image

ijemmao:
> That was the very first step I took.. view the error @ijemmao 
> 
> image
> 

@Ebugo can you share a screenshot of your code or push your code to show how this error is coming up exactly?

Ebugo:
Okay, here @ijemmao 

image

ijemmao (changes requested):

ijemmao:
> Okay, here @ijemmao
> 
> image

you'll probably want to extend the `ProcessEnv` interface by doing something like `interface ProcessEnv | 'build'` that way the interface type associated with `NODE_ENV` can handle the `'build'` value

Ebugo:
Didn't work for me.. do I push up so you try it out? @ijemmao

Ebugo:
I just pushed @ijemmao

ijemmao:
> I just pushed @ijemmao 
> 

I'll take a look on my computer later today, sorry for the back and forth!

.eslintrc.js
Ebugo: I added the overrides because the app is yet to go full on typescript, so that section takes care of the linting for the ts files.There'd be errors if only the ts linting is used since most of the app is still common js.

.eslintrc.js
ijemmao: but the `overrides` section shouldn't be used - rather all these unique additions that you made should be merged in the original section of this file

.eslintrc.js
ijemmao: you can have both eslint and ts linting running at the same time, we use that for our other open source project - i don't think it's the best to duplicate entries in this file if possible

ijemmao (changes requested):

ijemmao (dismissed):

PR #641, opened 2023-06-01
627 create developer information endpoint

davydocsurg:
## Describe your changes
created developer information endpoint

#627 

## How Has This Been Tested?
Tested with Postman

image


__tests__/developers.test.js
```diff
@@ -1,16 +1,5 @@
...
-  getWords,
-} from './shared/commands';
-import {
-  developerData,
-  malformedDeveloperData,
-  wordId,
-  exampleId,
-} from './__mocks__/documentData';
+import { createDeveloper, getDeveloper, getExample, getExamples, getWord, getWords } from './shared/commands';
+import { developerData, malformedDeveloperData, wordId, exampleId } from './__mocks__/documentData';
```
ijemmao, line 2: revert these changes

firebase.json
```diff
@@ -44,4 +44,4 @@
       "port": 8088
     }
   }
-}
+}
```
ijemmao, line 47: revert this change

src/middleware/validateApiKey.js
```diff
@@ -64,28 +24,16 @@ export default async (req, res, next) => {
     if (!apiKey && isDevelopment) {
       apiKey = FALLBACK_API_KEY;
     }
-    if (!apiKey) {
-      throw new Error('X-API-Key Header doesn\'t exist');
-    }
```
ijemmao, line 69: revert this change - we don't want to alter existing logic

I'm thinking this file shouldn't even be changed since there are changes to code used actively in production. feel free to reach out to me to get more clarification in you need some 🙇🏾‍♀️

ijemmao (changes requested):

__tests__/developers.test.js
```diff
@@ -19,13 +20,11 @@ describe('Developers', () => {
       expect(res.status).toEqual(200);
       expect(res.body.message).not.toEqual(undefined);
     });
-
     it('should throw an error while creating a new developer', async () => {
       const res = await createDeveloper(malformedDeveloperData);
       expect(res.status).toEqual(400);
       expect(res.body.error).not.toEqual(undefined);
     });
-
```
ijemmao, line 28: revert this line change

ijemmao (changes requested):

__tests__/developers.test.js
```diff
@@ -19,13 +20,11 @@ describe('Developers', () => {
       expect(res.status).toEqual(200);
       expect(res.body.message).not.toEqual(undefined);
     });
-
```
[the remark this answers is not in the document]
davydocsurg: @ijemmao, kindly explain. I don't understand this

firebase.json
ijemmao: @davydocsurg please revert this change - there shouldn't be a diff on this line

src/middleware/validateApiKey.js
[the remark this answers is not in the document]
ijemmao: it doesn't exist in your project version because you deleted that line - but if you take a look at this file but on the main branch in GitHub, you will see that the following is still in this file:

```js
if (!apiKey) {
  throw new Error('X-API-Key Header doesn\'t exist');
}
```

you don't want to delete this line because we want to catch the case where no API token is provided and we throw an error. if we don't, then we could allow the application to throw an even more obscure and detrimental issue down the line

so my recommendation is that you make as little changes to this file as possible to avoid introducing new possible bugs

ijemmao (changes requested):

src/middleware/validateApiKey.js
```diff
@@ -64,28 +20,20 @@ export default async (req, res, next) => {
     if (!apiKey && isDevelopment) {
       apiKey = FALLBACK_API_KEY;
     }
+
     if (!apiKey) {
-      throw new Error('X-API-Key Header doesn\'t exist');
```
ijemmao, line 68: this is going to fail linting - we don't use double quotes - you really don't need to make a change to this part of the code 😅

src/controllers/developers/utils.js
```diff
@@ -0,0 +1,61 @@
...
+  console.timeEnd('Finding developer account');
+  return developer;
+};
+
+export const checkDeveloperAPILimit = async (apiKey) => {
+  const developer = await findDeveloper(apiKey);
+
+  if (developer) {
+    if (developer.usage.count >= determineLimit(PROD_LIMIT)) {
+      throw new Error('You have exceeded your API limit. Please upgrade your plan.', { cause: 403 });
```
ijemmao, line 54: Let's not change error messages. This error message is misleading since developer currently don't have an option to upgrade their plan.

```suggestion
  throw new Error('You have exceeded your limit of requests for the day', { cause: 403 })
```

ijemmao (changes requested):
super close!

firebase.json
davydocsurg: Fixed, this should be Prettier's format changes

ijemmao (approved):

Issue #633, opened 2023-05-13
Add `.prettier` to the project


ijemmao:
## Background
Currently, developers need to manually format all their changed files to pass the linting GitHub action.

## Ideal Solution
Add `.prettier` and other prettier-related files to the project. Then update all the files in the codebase with the new `.prettier` instructions.

Chimise:
I would love to work on the issue

PR #645, opened 2023-06-05
Add prettier support to the project

Chimise:
## Describe your changes
Installed the prettier library and updated the lint-staged script to first run the prettier command to format the file before linting. Also updated the extends field in .eslintrc.js file adding the eslint-config-prettier plugin as the last plugin in the array to override configurations from other plugins and removed three eslint rules that conflict with prettier. The removed rules where detected using the eslint-config-prettier CLI. Added support for Typescript files

chore #633
## Motivation and Context
Currently, developers need to manually format all their changed files to pass the linting GitHub action #633

## How Has This Been Tested?
Tested the functionality manually, by committing an unformatted file

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,3 @@
-const os = require('os');
-
 module.exports = {
   env: {
     es2021: true,
@@ -7,7 +5,8 @@ module.exports = {
     jest: true,
   },
   ignorePatterns: ['**/functions/*.js'],
-  extends: ['next', 'airbnb', 'plugin:cypress/recommended'],
+  // Prettier plugin should be the last to always override preceding plugins
+  extends: ['next', 'airbnb', 'plugin:cypress/recommended', 'prettier'],
   parserOptions: {
     ecmaVersion: 12,
     sourceType: 'module',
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -1,13 +1,12 @@
...
 module.exports = {
   env: {
     es2021: true,
     node: true,
     jest: true,
   },
   ignorePatterns: ['**/functions/*.js'],
-  extends: ['next', 'airbnb', 'plugin:cypress/recommended'],
+  // Prettier plugin should be the last to always override preceding plugins
+  extends: ['next', 'airbnb', 'plugin:cypress/recommended', 'prettier'],
```
ijemmao, line 9: are we sure we want prettier to be last? that means that if there's a rule that only exists within prettier then the entire project could be changed?

package.json
```diff
@@ -57,8 +57,8 @@
     }
   },
   "lint-staged": {
-    "*.js": [
-      "./node_modules/.bin/eslint"
+    "*.{js,ts}": [
+      "./node_modules/.bin/prettier --write", "./node_modules/.bin/eslint"
```
ijemmao, line 61: i don't think we want prettier to write our changes when we're about to commit, since that introduces changes the developer didn't have time to review?

instead, we should introduce a `.prettierrc` file that defines our rules, and then each developer can set up their VScode to format on save so they can see their changes before committing

Chimise: A prettierrc.json file which handles that already exist in the project, prettier only reformats the code during commit to ensure all committed codes use the same formatting. I thought you wanted to enforce a consistent code style throughout the project, maybe I misunderstood the issue.

ijemmao (approved):

Issue #661, opened 2023-06-20
Convert `src/services`


ijemmao:
**Is your feature request related to a problem? Please describe.**
Convert `src/services` to TS

**Describe the solution you'd like**
Convert `src/services` to TS. There shouldn't be any logical changes to these files. Prettier will automatically format these files which is okay.

[Files to change](https://github.com/nkowaokwu/igbo_api/tree/master/src/services_

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Dhaxor:
Hello, I'm interested in taking this up.

ijemmao:
@Dhaxor assigned to you

PR #667, opened 2023-06-21
chore: convert the Services files to TS

Dhaxor:
## Describe your changes

Convert Services files to TS

#661

## Motivation and Context

Converting the entire project to TS

## How Has This Been Tested?

Locally


src/services/database.ts
```diff
@@ -24,8 +22,8 @@ export const createDbConnection = () => {
 };
 
 /* Closes current connection to MongoDB */
-export const disconnectDatabase = () => {
-  const db = mongoose.connection;
+export const disconnectDatabase = (): void => {
+  const db: mongoose.Connection = mongoose.connection;
```
ijemmao, line 26: nit: this doesn't need to be annotated since `mongoose.connection` already has the type `Connection`
```suggestion
  const db = mongoose.connection;
```

ijemmao (changes requested):
looking good! just one small comment

ijemmao:
great work 🚀

Issue #693, opened 2023-06-22
Sticky navbar is not working on mobile version


bhantsi:
**Describe the bug**
The sticky navbar on mobile disappears as the user scrolls down to various website sections.
**To Reproduce**
1. Go to the homepage on mobile
2. Scroll down to any section
3. See bug

**Expected behavior**
The navbar should remain fixed at the top (or any other specified position) of the webpage as the user scrolls. This ensures that the navigation menu is always accessible, regardless of the user's position on the page.
Bear in mind that the sticky navbar should adapt to different screen sizes and devices. It should be responsive and adjust its layout and appearance accordingly, ensuring optimal usability on mobile devices as well as desktops

**Screenshots**
![1687449622617](https://github.com/nkowaokwu/igbo_api/assets/26586086/34409e80-21bd-4ec4-b95c-72f8c4446932)

kingdanie:
@ijemmao this is the intended behaviour for mobile devices? yes if assign it to me, let me handle it.

ijemmao:
@kingdanie go for it!

ijemmao:
Closing since this issue since it doesn't seem relevant anymore. Feel free to reopen. Thank you for opening this ticket!

PR #703, opened 2023-06-26
fix: fixed navbar on all devices #693

kingdanie:
## Describe your changes
made navbar fixed for all devices

closed issue #693

## Motivation and Context
Make the navbar also fixed on mobile devices.

### DIFF
--- a/src/pages/components/Navbar/Navbar.js
+++ b/src/pages/components/Navbar/Navbar.js
@@ -11,9 +11,9 @@ const Navbar = ({ to, transparent }) => {
   const matchesLargeScreenQuery = useMediaQuery('(min-width:1024px)');
   return (
     <div
-      className={`flex items-center justify-between w-full py-5 lg:px-10
+      className={`flex fixed items-center justify-between w-full py-5 lg:px-10
       ${transparent ? 'transparent' : 'bg-white bg-opacity-75'} select-none`}
-      style={{ zIndex: 2, position: 'fixed' }}
+      style={{ zIndex: 2 }}
     >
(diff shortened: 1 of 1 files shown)


kingdanie:
@ijemmao  for some reason the class "fixed" was not giving the desired result, therefore as a hack I implemented it as an inline style . Hope this is ok?

ijemmao:
@kingdanie did you attempt to use the TAILWIND CSS fixed class? That should do the trick

kingdanie:
or could it have something to do with the tailwind config file?

it might not be related, but I just read from the tailwind docs that the **purge** property in the config file should be changed to **content**  when using v3. https://tailwindcss.com/docs/upgrade-guide#configure-content-sources

ijemmao:
> or could it have something to do with the tailwind config file?
> 
> it might not be related, but I just read from the tailwind docs that the **purge** property in the config file should be changed to **content** when using v3. https://tailwindcss.com/docs/upgrade-guide#configure-content-sources

that does seem related, but we do have the `content` field within `tailwind.config.js` so if anything the `purge` field should be ignored 😕 

did you attempt to run `yarn build` to build the project and then `yarn start`? that will build the project just like in production and you can see if the styles get applied there - if not, we can revert back to your initial approach and then create a follow up issue to address this

kingdanie:
yeah, that was exactly the issue I forgot to build after making the change. I however noticed it and added the tailwind class instead of the inline style in my last commit.

ijemmao:
@kingdanie for visual changes, can you please attach a video of this change working as expected on mobile?

kingdanie:
sure

kingdanie:
Screenshot 2023-07-01 at 15 24 20

PR #705, opened 2023-06-29
fix: detect os for the es lint line breaker rule

Ndohjapan:
## Describe your changes

To avoid encountering the line breaker error when trying to make a commit, I changed the eslint config to js in order to use the `os` module to detect the device os platform. With this, the eslint then chooses the line breaker rule to follow

N/A

## Motivation and Context
It prevented me from making commits cause the default was set to unix while I used a windows device.

## How Has This Been Tested?
After the eslint file was configured again, all files had the squiggly line removed at the end of each line regarding the line breaker rule.

image


ijemmao:
@Ndohjapan seems like this is breaking the tests?

Ndohjapan:
I will use github codespace to make the navbar changes in a unix environment.

Issue #704, opened 2023-06-29
Create `/login` endpoint


davydocsurg:
Create the Express endpoint and you don’t need to define the business logic of logging in - this should be `POST` request

PR #706, opened 2023-06-29
704 create login endpoint

davydocsurg:
## Describe your changes
created the login endpoint and controller

#704


src/controllers/auth/login.js
```diff
@@ -0,0 +1 @@
+export const login = () => {};
```
ijemmao, line 1: won't this hang the request since you aren't calling `res.send({})`?

__tests__/auth.test.js
```diff
@@ -0,0 +1,5 @@
+describe('Login', () => {
+  describe('Login endpoint', () => {
+    it('should hit the login endpoint', () => {});
```
ijemmao, line 3: it's not good practice to write empty `it` test blocks - there's no real need to add a test since the controller function doesn't have any logic to test

src/routers/router.js
```diff
@@ -11,6 +11,7 @@ import validateAdminApiKey from '../middleware/validateAdminApiKey';
 import attachRedisClient from '../middleware/attachRedisClient';
 import analytics from '../middleware/analytics';
 import authorization from '../middleware/authorization';
+import { login } from '../controllers/auth/login';
```
ijemmao, line 14: nit: can you move this import to be right below `getStats` so all the controller imports are near each other?

ijemmao (changes requested):
great work ✨

ijemmao (approved):

src/controllers/auth/login.js
```diff
@@ -0,0 +1,11 @@
+export const login = (res) => {
+  try {
+    return res.status(200).send({
+      message: 'Logging in...',
+    });
+  } catch (error) {
+    return res.status(400).send({
+      error: error.message,
+    });
```
ijemmao, line 9: @davydocsurg now that you've added logic, you'll need write a test for the 200 error response

ijemmao, line 9: @davydocsurg we have an errorHandler express middleware, so instead of short-circuiting the chain of methods in this requests, you want to use the `next` method: https://expressjs.com/en/guide/writing-middleware.html

src/controllers/auth/login.js
```diff
@@ -0,0 +1,11 @@
+export const login = (res) => {
```
ijemmao, line 1: i'm surprised this is working since the first argument provided in an Express route callback function is `req` and not `res`
```suggestion
export const login = (req, res, next) => {
```

ijemmao (changes requested):

src/controllers/auth/login.js
ijemmao: @davydocsurg just want to call out this comment again

ijemmao (approved):

Issue #702, opened 2023-06-26
Implement the Navbar Component using the latest Figma design.


davydocsurg:
Please find the Figma design for the Navbar Component at the following link: [Figma Design for Navbar](https://www.figma.com/file/KEdhBREtlZKSXVMw95lZTg/Igbo-API-Editor-Platform)

davydocsurg:
Yes, it is

PR #707, opened 2023-06-30
feat: created the developer nav bar

Ndohjapan:
## Describe your changes
Created the Navbar for developer dashbaord

https://github.com/nkowaokwu/igbo_api/issues/702

image
image


.yarn/releases/yarn-1.22.19.cjs
ijemmao, line 1: @Ndohjapan please delete this file

.yarnrc
ijemmao, line 1: @Ndohjapan please delete this file

src/pages/components/Navbar/DeveloperNavbar.js
```diff
@@ -0,0 +1,81 @@
+import React, { useState } from 'react';
+import { Dialog } from '@headlessui/react';
+import { Bars3Icon, XMarkIcon } from '@heroicons/react/24/outline';
```
ijemmao, line 3: we don't need these packages since we have icons from material UI or chakra - please uninstall these packages

src/pages/components/Navbar/DeveloperNavbar.js
```diff
@@ -0,0 +1,81 @@
...
+  { name: 'About', href: '#' },
+  { name: 'Register API Key', href: '#' },
+  { name: 'Docs', href: '#' },
+];
+
+export default function NewHeader() {
+  const [mobileMenuOpen, setMobileMenuOpen] = useState(false);
+
+  return (
+    <div className="bg-white">
```
ijemmao, line 17: we use ChakraUI for this project - please replace all div, a, h1, etc. with their ChakraUI equivalents (Box, Link, Heading, etc.)

src/pages/components/Navbar/DeveloperNavbar.js
```diff
@@ -0,0 +1,81 @@
...
+      <header className="absolute inset-x-0 top-0 z-50">
```
ijemmao, line 18: it's not typical to use the `header` tag - please refer to the navigation bar that's used on the homepage for inspiration

src/pages/components/Navbar/DeveloperNavbar.js
```diff
@@ -0,0 +1,81 @@
...
+const navigation = [
+  { name: 'Features', href: '#' },
+  { name: 'Try it Out', href: '#' },
+  { name: 'About', href: '#' },
+  { name: 'Register API Key', href: '#' },
+  { name: 'Docs', href: '#' },
+];
+
+export default function NewHeader() {
+  const [mobileMenuOpen, setMobileMenuOpen] = useState(false);
```
ijemmao, line 14:
```suggestion
  const [isMobileMenuOpen, setIsMobileMenuOpen] = useState(false);
```

ijemmao (changes requested):
great start~

src/pages/components/Navbar/DeveloperNavbar.js
```diff
@@ -0,0 +1,81 @@
+import {
```
ijemmao, line 1: question: why are you creating a brand new navbar? these are the same options that we have in our existing navbar - couldn't we reuse the navbar that we already have?

Ndohjapan: I actually wanted to use the existing nav bar.

However the Figma design is a bit different from it and I thought it was intentional from the design to make that of the developer dashboard different from the main nav bar.

ijemmao: hmm, let's confirm with David if we can update the Figma designs so they match our current navbar, or if he wants to update our existing navbar to match his newer designs

i think having two navbars at right now might be a bit much for us to maintain!

PR #723, opened 2023-07-26
chore: convert src/middleware files to TS

### LINKED ISSUE #670: Convert `src/middleware` to TS
Convert `src/middleware` to TS

Convert [`src/middleware`](https://github.com/nkowaokwu/igbo_api/tree/master/src/middleware) to TS

Avoid making logic changes to these files. Some of these files will need new types and interfaces, add them wherever necessary.

Prettier will format these files which is okay.

N/A

**Additional context**
N/A

chidexebere:
## Describe your changes
- This updates the `.eslintrc.js` to allow dangling underscores in identifiers
- This converts `.js` files in `src/middleware` to `.ts`

Closed #670 

## Motivation and Context
Migrating project to TypeScript

## How Has This Been Tested?
Locally

N/A


.eslintrc.js
```diff
@@ -62,6 +62,7 @@ module.exports = {
         ],
         '@typescript-eslint/no-empty-function': ['warn'],
         'prefer-const': 'error',
+        'no-underscore-dangle': 'off',
```
ijemmao, line 65: why did you introduce this change? specifically, which line of code was getting affected by this rule when it was enabled?

src/middleware/analytics.ts
```diff
@@ -0,0 +1,85 @@
...
+        });
+    }
+  });
+
+export default async (req, _, next) => {
+  try {
+    const { method } = req;
+    const developerAPIKey = req.headers['X-API-Key'] || req.headers['x-api-key'];
+    const { keyword } = req.query;
+    const { pathname } = req._parsedUrl;
```
chidexebere, line 72: .... added the `'no-underscore-dangle': 'off',` eslint rule because of this line

ijemmao: i don't think we should change the eslint rules for one case

let's add `// ts-expect-error eslint(no-underscore-dangle) so that we can address this error in a follow ticket
```suggestion
    // ts-expect-error eslint(no-underscore-dangle)
    const { pathname } = req._parsedUrl;
```

I did a little investigating and when I add `req: Request` it looks like `_parsedUrl` doesn't live on the request object, so this is a clear sign that we actually want to stop using `_parsedUrl` and change it to a typed equivalent

but for this PR, let's just add the ts-expect-error and you can take on the follow up ticket to change this variable (along side add types function arguments)

ijemmao (changes requested):

src/middleware/analytics.ts
chidexebere: For some reason, `ts-expect-error eslint(no-underscore-dangle)` does not work.
The compiler still throws an error.

But this works if added on that line:
/* eslint no-underscore-dangle: 0 */

Should I go ahead and make the changes?

ijemmao: Sorry, it's not a TS error rather it's a ESLint error

You want to use the following comment above the line of code that has the error:
`// eslint-disable-next-line no-underscore-dangle`

ijemmao (approved):

PR #729, opened 2023-08-09
chore: Use the correct argument types in src/middleware files

### LINKED ISSUE #728: Use the correct argument types in `src/middleware`
Arguments in functions in files under the `src/middleware` folder do not have their correct types.

- Use the correct argument types for functions in these files.
- Introduce new types or interfaces to correctly type these files.
- Don't introduce any logical changes to these files.

N/A

**Additional context**
N/A

chidexebere:
## Describe your changes

- This adds the correct type for arguments in the functions in `src/middleware` files
- This creates new types and interfaces where needed
- This uses the `any` type in few arguments where their type is dependent on files outside this folder or can not be easily determined

Closed #728

## Motivation and Context
Migrating entire project to TypeScript

## How Has This Been Tested?
Locally

N/A

### DIFF
--- a/src/middleware/analytics.ts
+++ b/src/middleware/analytics.ts
@@ -2,14 +2,14 @@ import axios from 'axios';
 import { NextFunction, Request } from 'express';
 import { GA_TRACKING_ID, GA_API_SECRET, GA_URL, DEBUG_GA_URL, isProduction as isProductionConfig } from '../config';
 
-interface Props {
+interface TrackingEvent {
   clientIdentifier: string | string[] | undefined;
   category: string;
   action: string;
   keyword: any;
 }
 
-const trackEvent = ({ clientIdentifier, category, action, keyword }: Props) =>
+const trackEvent = ({ clientIdentifier, category, action, keyword }: TrackingEvent) =>
   new Promise((resolve, reject) => {
     const params = {
       measurement_id: GA_TRACKING_ID,
@@ -72,7 +72,7 @@ const trackEvent = ({ clientIdentifier, category, action, keyword }: Props) =>
     }
   });
 
-export default async (req: Request, next: NextFunction) => {
+export default async (req: Request, _, next: NextFunction) => {
   try {
     const { method } = req;
     const developerAPIKey = req.headers['X-API-Key'] || req.headers['x-api-key'];
--- a/src/middleware/analytics.ts
+++ b/src/middleware/analytics.ts
@@ -1,7 +1,15 @@
 import axios from 'axios';
+import { NextFunction, Request } from 'express';
 import { GA_TRACKING_ID, GA_API_SECRET, GA_URL, DEBUG_GA_URL, isProduction as isProductionConfig } from '../config';
 
-const trackEvent = ({ clientIdentifier, category, action, keyword }) =>
+interface Props {
+  clientIdentifier: string | string[] | undefined;
+  category: string;
+  action: string;
+  keyword: any;
+}
+
+const trackEvent = ({ clientIdentifier, category, action, keyword }: Props) =>
   new Promise((resolve, reject) => {
     const params = {
       measurement_id: GA_TRACKING_ID,
@@ -64,13 +72,12 @@ const trackEvent = ({ clientIdentifier, category, action, keyword }) =>
     }
   });
 
-export default async (req, _, next) => {
+export default async (req: Request, next: NextFunction) => {
   try {
     const { method } = req;
     const developerAPIKey = req.headers['X-API-Key'] || req.headers['x-api-key'];
     const { keyword } = req.query;
-    // eslint-disable-next-line no-underscore-dangle
-    const { pathname } = req._parsedUrl;
+    const { pathname } = req.params;
 
     await trackEvent({
       clientIdentifier: developerAPIKey || 'anon_client_id',
--- a/src/middleware/attachRedisClient.ts
+++ b/src/middleware/attachRedisClient.ts
@@ -11,7 +11,7 @@ const afterResponse = (redisClient: RedisClientType) => {
     if (redisClient) {
       redisClient.quit();
     }
-  } catch (err: any) {
+  } catch (err) {
     console.log(`Error with closing redis: ${err.message}`);
   }
 };
--- a/src/middleware/attachRedisClient.ts
+++ b/src/middleware/attachRedisClient.ts
@@ -1,12 +1,17 @@
-import { createClient } from 'redis';
+import { createClient, RedisClientType } from 'redis';
+import { NextFunction, Request, Response } from 'express';
 import { REDIS_HOST, REDIS_PORT, REDIS_URL, REDIS_USERNAME, REDIS_PASSWORD } from '../config';
 
-const afterResponse = (redisClient) => {
+interface RedisClientRequest extends Request {
+  redisClient: RedisClientType;
+}
+
+const afterResponse = (redisClient: RedisClientType) => {
   try {
     if (redisClient) {
       redisClient.quit();
     }
-  } catch (err) {
+  } catch (err: any) {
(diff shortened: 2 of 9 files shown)


src/middleware/analytics.ts
```diff
@@ -64,13 +72,12 @@ const trackEvent = ({ clientIdentifier, category, action, keyword }) =>
     }
   });
 
-export default async (req, _, next) => {
+export default async (req: Request, next: NextFunction) => {
```
ijemmao, line 75: this will through an error since Express is expecting at least three arguments for endpoint handlers - please include `_`

src/middleware/attachRedisClient.ts
```diff
@@ -1,12 +1,17 @@
...
+  redisClient: RedisClientType;
+}
+
+const afterResponse = (redisClient: RedisClientType) => {
   try {
     if (redisClient) {
       redisClient.quit();
     }
-  } catch (err) {
+  } catch (err: any) {
```
ijemmao, line 14: we don't want to use `any`, rather it should be a defined or expected type like `Error`

src/middleware/logger.ts
```diff
@@ -1,6 +1,7 @@
 import chalk from 'chalk';
+import { NextFunction, Request } from 'express';
 
-export default (req, _, next) => {
+export default (req: Request, _: any, next: NextFunction) => {
```
ijemmao, line 4: we don't want to use `any` - rather if we need to specify types (which I don't think we need to?) we should provide `Response`

src/middleware/errorHandler.ts
```diff
@@ -1,5 +1,6 @@
-// eslint-disable-next-line
-export default (err, req, res, next) => {
+import { Response } from 'express';
+
+export default (err: any, res: Response) => {
```
ijemmao, line 3: again, we don't want to use `any` and we don't want to remove all these arguments - please add all these arguments back into this function as it will break the error handler

src/middleware/validId.ts
```diff
@@ -1,14 +1,15 @@
...
+export default (req: Request, res: Response, next: NextFunction) => {
   try {
     const { id } = req.params;
     if (!mongoose.Types.ObjectId.isValid(id)) {
       res.status(400);
       return res.send({ error: 'Provided an invalid id' });
     }
     return next();
-  } catch (err) {
+  } catch (err: any) {
```
ijemmao, line 12: we don't want to use `any`

src/middleware/validateAdminApiKey.ts
```diff
@@ -9,7 +10,7 @@ export default async (req, res, next) => {
     }
 
     return next();
-  } catch (err) {
+  } catch (err: any) {
```
ijemmao, line 13: we don't want to use `any`

src/middleware/validateApiKey.ts
```diff
@@ -1,20 +1,21 @@
...
+import { NextFunction, Request, Response } from 'express';
 import { developerSchema } from '../models/Developer';
 import { MAIN_KEY, isTest, isDevelopment, isProduction } from '../config';
 import { createDbConnection } from '../services/database';
 
 const PROD_LIMIT = 2500;
 const FALLBACK_API_KEY = 'fallback_api_key';
 
-const determineLimit = (apiLimit) => (isTest ? apiLimit || PROD_LIMIT : PROD_LIMIT);
+const determineLimit = (apiLimit: any) => (isTest ? apiLimit || PROD_LIMIT : PROD_LIMIT);
```
ijemmao, line 10: we don't want to use `any`

src/middleware/validateApiKey.ts
```diff
@@ -1,20 +1,21 @@
...
 
-const isSameDate = (first, second) =>
+const isSameDate = (first: Date, second: Date) =>
   first.getFullYear() === second.getFullYear() &&
   first.getMonth() === second.getMonth() &&
   first.getDate() === second.getDate();
 
 /* Increments usage count and updates usage date */
-const handleDeveloperUsage = async (developer) => {
+const handleDeveloperUsage = async (developer: any) => {
```
ijemmao, line 18: we don't want to use `any`

src/middleware/validateApiKey.ts
```diff
@@ -42,9 +43,9 @@ const findDeveloper = async (apiKey) => {
   // This logic attempts to find the developer document and update it
   // with the API token
   const developers = await Developer.find({});
-  developer = developers.find((dev) => compareSync(apiKey, dev.apiKey));
+  developer = developers.find((dev) => compareSync(apiKey as string, dev.apiKey))!;
```
ijemmao, line 46: we don't want to say `as string`, `apiKey` should be typed already

src/middleware/validateApiKey.ts
```diff
@@ -42,9 +43,9 @@ const findDeveloper = async (apiKey) => {
...
   if (developer) {
-    developer.apiKey = apiKey;
+    developer.apiKey = apiKey as string;
```
ijemmao, line 48: we don't want to say `as string`, `apiKey` should be typed already

src/middleware/validateApiKey.ts
```diff
@@ -90,7 +95,7 @@ export default async (req, res, next) => {
 
     res.status(401);
     return res.send({ error: 'Your API key is invalid' });
-  } catch (err) {
+  } catch (err: any) {
```
ijemmao, line 98: we don't want to use `any`

src/middleware/validateApiKey.ts
```diff
@@ -79,7 +84,7 @@ export default async (req, res, next) => {
 
     const developer = await findDeveloper(apiKey);
 
-    if (developer) {
+    if (developer && developer.usage) {
```
ijemmao, line 87: there's no need to check to see if a developer has `usage` on it since we have require developer to always have it - lets revert this line

src/middleware/analytics.ts
```diff
@@ -1,7 +1,15 @@
 import axios from 'axios';
+import { NextFunction, Request } from 'express';
 import { GA_TRACKING_ID, GA_API_SECRET, GA_URL, DEBUG_GA_URL, isProduction as isProductionConfig } from '../config';
 
-const trackEvent = ({ clientIdentifier, category, action, keyword }) =>
+interface Props {
```
ijemmao, line 5: `Props` is more for React components
```suggestion
interface TrackingEvent {
```

src/middleware/attachRedisClient.ts
```diff
@@ -44,6 +49,6 @@ export default async (req, res, next) => {
   res.on('finish', afterResponse);
   res.on('close', afterResponse);
 
-  req.redisClient = redisClient;
+  req.redisClient = redisClient as RedisClientType;
```
ijemmao, line 52: rather than doing `as RedisClientType` here we want to annotate the `redisClient` when it's getting created on line 20

so we should do `const redisClient: RedisClientType = ...` if that doesn't work then you can cast the `redisClient` by using `as RedisClientType` on line 41.

```suggestion
  req.redisClient = redisClient;
```

src/middleware/validateApiKey.ts
ijemmao, line 46: please remove the bang (`!`) at the end of this line since we check on line 47 if the developer exists or not

ijemmao (changes requested):
really great work!

src/middleware/attachRedisClient.ts
chidexebere: Yes, your suggestion is the right way to do it but it does not work "When Typescript is strict: true" because of some issues with RedisClientType. You can read [here](https://github.com/redis/node-redis/issues/1865). But will make the change

ijemmao: makes sense, thanks for the link

Issue #725, opened 2023-07-27
Add Rate limit to all endpoint


Vickysomtee:
**Is your feature request related to a problem? Please describe.**
Currently only the endpoint for creating developer accounts is rate limted which leaves other endpoints to  vulnerable to malicious actors

**Describe the solution you'd like**
Rate limit other endpoint

tohuw:
This is now implemented in #836 (a revival of the stalled #733): `router.use()`-mounted rate limiting on both the v1 and v2 routers, an IP allowlist via `RATE_LIMIT_ALLOWED_IPS` for trusted callers, a tunable ceiling via `RATE_LIMIT_MAX_REQUESTS`, and 6 passing unit tests. This issue can close when that PR lands.

---
*This comment was created by Claude Fable 5 on behalf of [@tohuw](https://github.com/tohuw).*

PR #733, opened 2023-08-13
feat: add rate limiter

Vickysomtee:
## Describe your changes
This PR adds rate limiters to all developers endpoints

 #725 

## Motivation and Context
See issue - #725 

## How Has This Been Tested?
Testing locally with Postman

Screenshot 2023-08-13 at 12 59 42


src/routers/router.ts
```diff
@@ -11,26 +10,20 @@ import validateApiKey from '../middleware/validateApiKey';
...
-});
-
 // Google Analytics
 router.use(analytics);
 
-router.get('/words', validateApiKey, attachRedisClient, getWords);
-router.get('/words/:id', validateApiKey, validId, attachRedisClient, getWord);
-router.get('/examples', validateApiKey, attachRedisClient, getExamples);
-router.get('/examples/:id', validateApiKey, validId, attachRedisClient, getExample);
+router.get('/words', rateLimiter, validateApiKey, attachRedisClient, getWords);
```
ijemmao, line 20: instead of using `raterLimiter` within each route you can add the follow line before the first `router.get`

```ts
router.us(rateLimiter)
```

ref: https://expressjs.com/en/4x/api.html#app.use

src/routers/routerV2.ts
```diff
@@ -5,13 +5,14 @@ import validId from '../middleware/validId';
...
 import attachRedisClient from '../middleware/attachRedisClient';
+import { rateLimiter } from '../middleware/rateLimiter';
 
 const routerV2 = express.Router();
 
-routerV2.get('/words', analytics, validateApiKey, attachRedisClient, getWords);
-routerV2.get('/words/:id', analytics, validateApiKey, validId, attachRedisClient, getWord);
-routerV2.get('/examples', analytics, validateApiKey, attachRedisClient, getExamples);
-routerV2.get('/examples/:id', analytics, validateApiKey, validId, attachRedisClient, getExample);
+routerV2.get('/words', rateLimiter, analytics, validateApiKey, attachRedisClient, getWords);
```
ijemmao, line 12: same comment here

ijemmao (changes requested):
this is a great start! can you also add unit tests that test this functionality? specifically:
* a test for checking to see if rateLimiting works as expected for router v1
* a test for checking to see if rateLimiting works as expected for router v2

src/routers/router.ts
```diff
@@ -11,26 +10,19 @@ import validateApiKey from '../middleware/validateApiKey';
...
 router.get('/words', validateApiKey, attachRedisClient, getWords);
 router.get('/words/:id', validateApiKey, validId, attachRedisClient, getWord);
 router.get('/examples', validateApiKey, attachRedisClient, getExamples);
 router.get('/examples/:id', validateApiKey, validId, attachRedisClient, getExample);
 
-router.post('/developers', createDeveloperLimiter, validateDeveloperBody, postDeveloper);
+router.post('/developers', validateDeveloperBody, postDeveloper);
 
-router.get('/stats', validateAdminApiKey, attachRedisClient, getStats);
+router.get('/stats', rateLimiter, validateAdminApiKey, attachRedisClient, getStats);
```
ijemmao, line 26: why add `rateLimit` specifically to this route? it's already applied to this router

src/middleware/rateLimiter.ts
```diff
@@ -0,0 +1,10 @@
+import rateLimit from 'express-rate-limit';
+import { Express } from '../types';
+
+export const rateLimiter: Express.MiddleWare = rateLimit({
```
ijemmao, line 4: Is there a possible way we can allow specific IP addresses to not get stopped by the rate limiter? For instance, [Nkọwa okwu](https://nkowaokwu.com) uses the Igbo API and shouldn't be rate limited the same where an individual user should be rate limited

ijemmao (changes requested):
looking really solid! just a few questions

src/middleware/rateLimiter.ts
Vickysomtee: Yes. I think this is posssible. We can have the IP addresses in a whitelist file. In the middleware,  we read the file to get these addresses then call the next function if the incoming request IP is same as any of the whitelisted IP address.

Does this make sense?

src/routers/router.ts
Vickysomtee: A little error

src/middleware/rateLimiter.ts
ijemmao: yeah, that makes a lot of sense - do you know of any guides/tutorials/articles that walk through this process specifically for rate limiting node js / express apps?

i believe you can get a shortlist of valid GCP server IP addresses that we can whitelist since we are deploying this project on GCP

ijemmao (changes requested):

tohuw:
@Vickysomtee this stalled on conflicts after the review, so to save you the rebase archaeology I've rebuilt it on current master in #836 with you kept as commit co-author — it implements everything from the review: `router.use()` mounting on both v1 and v2, the IP-allowlist bypass @ijemmao asked about (`RATE_LIMIT_ALLOWED_IPS` env var), and the requested unit tests (6, all passing).

If you'd rather land it from this PR instead, feel free to cherry-pick the commit from [`tohuw:feat/rate-limiter-revival`](https://github.com/tohuw/igbo_api/tree/feat/rate-limiter-revival) onto your branch and I'll close #836 — no ego in this, just trying to get your change over the line.

---
*This comment was created by Claude Fable 5 on behalf of [@tohuw](https://github.com/tohuw).*

PR #743, opened 2023-08-30
docs: correct api key header name in docs Fixes #742

### LINKED ISSUE #742: Feedback for API Key
The API Key header name is said to be `X-API-Header` in the docs, but `X-API-Key` is what really works. You might want to amend the documentation. Thank you for this fantastic tool though.

Emmo00:
## Describe your changes
Changed the header name for the API key on the docs from `X-API-Header` to `X-API-Key`, which is what the app actually looks for.

Closed #742 

## Motivation and Context
Gives a better and easier experience to future developers that use the docs.

## How Has This Been Tested?
This should not be a breaking change to the application.

### DIFF
--- a/src/pages/docs/getting-started.mdx
+++ b/src/pages/docs/getting-started.mdx
@@ -48,7 +48,7 @@ Once you've configured your project, you can start the Igbo API dev server by ru
 yarn dev
 ```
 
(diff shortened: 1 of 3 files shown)


ijemmao (approved):
lgtm!

Issue #671, opened 2023-06-21
Convert `src/controllers/utils` to TS


ijemmao:
**Is your feature request related to a problem? Please describe.**
Convert `src/controllers/utils` to TS

**Describe the solution you'd like**
Conver [`src/controllers/utils`](https://github.com/nkowaokwu/igbo_api/tree/master/src/controllers/utils) to TS

Avoid making logical changes to these files. Some of these files will need new types or interfaces, add them whenever necessary. Prettier will format these files which is okay.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Harshal662:
There are 3 files in .js format so all the 3 files should be converted to TS?

ijemmao:
@Harshal662 that's correct

i assigned to both of you @Dhaxor @Harshal662

PR #749, opened 2023-09-09
Convert controller util files to typescript

zanio:
## Describe your changes
 * Convert top level JS in /controllers/utils to TS
 * install relevant type definitions for modules that don't ship their own types
 * Create new global and file scoped types where necessary
 * Relocated types that are needed by other external modules into the global types module
 * Minimal but inevitable refactoring in order to make files strongly typed and also compile, for example, I wrapped this existing code
 ```TS
// src/controllers/utils/minimizeWords.ts
    minimizedDefinition = omit(minimizedDefinition as Partial<Definition>, [
                'label',
                'igboDefinitions',
                '_id',
                'id',
              ]);
              if (!minimizedDefinition.nsibidi) {
                minimizedDefinition = omit(minimizedDefinition, ['nsibidi']);
              }
```

inside a typeof check condition below:
 ```TS
// src/controllers/utils/minimizeWords.ts
   if (typeof definition === 'object') {
              ...existing bits
            }
```
* gitignored  my ide generated folder for the project ( I use intellij, and it generates `.idea` folder )

Closed [#671](https://github.com/nkowaokwu/igbo_api/issues/671)

## How Has This Been Tested?

Regression Test was done, 
* Ran `yarn build` to see that typescript compiles as expected
* Ran `yarn test` to see that the little refactor didn't introduce any defects
* Ran `yarn dev` to test that portion worked on

Screenshot 2023-09-09 at 07 57 31

### DIFF
--- a/src/APIs/RedisAPI.ts
+++ b/src/APIs/RedisAPI.ts
@@ -6,6 +6,7 @@ import Version from '../shared/constants/Version';
 import { Word } from '../types';
 import { ExampleResponseData } from '../controllers/types';
 import { redisClient as defaultRedisClient } from '../middleware/attachRedisClient';
+import { PartialWordType } from '../types/word';
 
 type RedisClient = {
   get: (value: string) => void;
@@ -39,7 +40,7 @@ export const setCachedWords = async ({
   version: Version;
 }) => {
   const updatedData = assign(data);
-  updatedData.words = minimizeWords(data.words, version);
+  updatedData.words = minimizeWords(data.words as PartialWordType[], version);
   if (!redisClient.isFake) {
     await redisClient.set(key, JSON.stringify(updatedData), { EX: REDIS_CACHE_EXPIRATION });
   }
--- a/src/APIs/RedisAPI.ts
+++ b/src/APIs/RedisAPI.ts
@@ -35,12 +35,15 @@ export const setCachedWords = async ({
   version,
 }: {
   key: string;
-  data: any;
+  data: {
+    words: PartialWordType[] | any;
+    contentLength: number;
+  };
   redisClient: RedisClient | undefined;
   version: Version;
 }) => {
   const updatedData = assign(data);
-  updatedData.words = minimizeWords(data.words as PartialWordType[], version);
+  updatedData.words = minimizeWords(data.words, version);
   if (!redisClient.isFake) {
     await redisClient.set(key, JSON.stringify(updatedData), { EX: REDIS_CACHE_EXPIRATION });
   }
--- a/src/controllers/utils/index.ts
+++ b/src/controllers/utils/index.ts
@@ -170,7 +170,7 @@ export const handleQueries = async ({
             }),
             ['wordReg']
           );
-          const keyWord: Partial<Keyword> = {
+          const keyWord: Keyword = {
             text,
             wordClass,
             regex: pickedRegex,
--- a/src/controllers/utils/index.ts
+++ b/src/controllers/utils/index.ts
@@ -13,8 +13,8 @@ import WordClass from '../../shared/constants/WordClass';
 import { getAllCachedVerbsAndSuffixes, setAllCachedVerbsAndSuffixes } from '../../APIs/RedisAPI';
 import convertToSkipAndLimit from './convertToSkipAndLimit';
 import parseRange from './parseRange';
-import { WordData } from './types';
-import { WithPronunciation } from '../types';
+import { WordData, Keyword } from './types';
+import { Filters, WithPronunciation } from '../types';
 
 const createSimpleRegExp = (keywords: { text: string }[]) => ({
   wordReg: new RegExp(
@@ -246,7 +246,7 @@ export const handleQueries = async ({
     examples,
     resolve,
   };
-  const filters = {
+  const filters: Filters = {
     ...(tags?.length ? { tags: { $in: tags } } : {}),
     ...(wordClasses?.length ? { 'definitions.wordClass': { $in: wordClasses } } : {}),
   };
@@ -256,7 +256,7 @@ export const handleQueries = async ({
     id,
     version,
     searchWord,
-    keywords,
+    keywords: keywords as Keyword[],
     regex,
     page,
     skip,
--- a/src/controllers/utils/index.ts
+++ b/src/controllers/utils/index.ts
@@ -162,17 +162,21 @@ export const handleQueries = async ({
   console.log(`Search word: ${searchWord}`);
   let keywords =
     version === Version.VERSION_2 && searchWord
-      ? expandVerb(searchWord, allVerbsAndSuffixes).map(({ text, wordClass }) => ({
-          text,
-          wordClass,
-          regex: pick(
+      ? expandVerb(searchWord, allVerbsAndSuffixes).map(({ text, wordClass }) => {
+          const pickedRegex = pick(
             constructRegexQuery({
               isUsingMainKey,
               keywords: [{ text }],
             }),
             ['wordReg']
-          ),
-        }))
+          );
+          const keyWord: Partial<Keyword> = {
+            text,
+            wordClass,
+            regex: pickedRegex,
+          };
+          return keyWord;
+        })
       : [];
   // Attempt to breakdown as noun if there is no breakdown as verb
   if (!keywords.length && searchWord) {
@@ -256,7 +260,7 @@ export const handleQueries = async ({
     id,
     version,
     searchWord,
-    keywords: keywords as Keyword[],
+    keywords,
     regex,
     page,
     skip,
--- a/src/controllers/utils/minimizeWords.ts
+++ b/src/controllers/utils/minimizeWords.ts
@@ -34,7 +34,7 @@ const minimizeWords = (words: PartialWordType[], version: Version) => {
             }
             return minimizedDefinition;
           }) as Partial<Definition>[] | string[])
-        : (minimizedWord.definitions as undefined);
+        : minimizedWord.definitions;
     if (!minimizedWord.variations?.length) {
       minimizedWord = omit(minimizedWord, ['variations']);
     }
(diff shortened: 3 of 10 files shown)


src/controllers/utils/minimizeWords.ts
```diff
@@ -49,13 +67,14 @@ const minimizeWords = (words, version) => {
...
     if (version === Version.VERSION_2 && minimizedWord.dialects?.length) {
-      minimizedWord.dialects = minimizedWord.dialects?.map((dialect) => {
-        let minimizedDialect = omit(dialect, ['variations', 'id', '_id']);
-        if (!minimizedDialect.pronunciation) {
-          minimizedDialect = omit(minimizedDialect, ['pronunciation']);
-        }
-        return minimizedDialect;
-      });
+      if (Array.isArray(minimizedWord.dialects))
+        minimizedWord.dialects = minimizedWord.dialects?.map((dialect) => {
```
ijemmao, line 71: remove the `?.` if you have already checked to see if `minimizedWord.dialects` is an array

src/controllers/utils/types.ts
```diff
@@ -28,3 +29,11 @@ export type Solution = {
 };
 
 export type TopSolution = { solution: Solution; metaData: Meta };
+
+export type Regex = SearchRegExp;
```
ijemmao, line 33: why are you re-exporting `Regex` when you could use `SearchRegExp` directly?

ijemmao (changes requested):
super great work! just two comments

src/controllers/utils/types.ts
zanio: Oh, this was an oversight, I initially had an extra typed field joined to SearchRegExp. I'll remove the Regex type

src/controllers/utils/index.ts
```diff
@@ -256,7 +256,7 @@ export const handleQueries = async ({
     id,
     version,
     searchWord,
-    keywords,
+    keywords: keywords as Keyword[],
```
ijemmao, line 259: why are we casting keywords as `Keyword[]` here? shouldn't the other methods that assign a value to `keywords` automatically return the `Keyword` type?

src/controllers/utils/minimizeWords.ts
```diff
@@ -1,29 +1,47 @@
...
+              ]);
+              if (!minimizedDefinition.nsibidi) {
+                minimizedDefinition = omit(minimizedDefinition, ['nsibidi']);
+              }
             }
             return minimizedDefinition;
-          })
-        : minimizedWord.definitions;
+          }) as Partial<Definition>[] | string[])
+        : (minimizedWord.definitions as undefined);
```
ijemmao, line 37: why are you casting `definitions` as `undefined`? it's not expected that `minimizedWord.definitions` is `undefined` if we using Version 1 of the API

src/APIs/RedisAPI.ts
```diff
@@ -39,7 +40,7 @@ export const setCachedWords = async ({
   version: Version;
 }) => {
   const updatedData = assign(data);
-  updatedData.words = minimizeWords(data.words, version);
+  updatedData.words = minimizeWords(data.words as PartialWordType[], version);
```
ijemmao, line 43: instead of casting `data.words` to `PartialWordType[]` on this line, we should actually change the `data: any` type on line 38 to expect its type to be `PartialWordType[]`

ijemmao (changes requested):

src/controllers/utils/searchWordUsingIgbo.ts
```diff
@@ -8,11 +8,12 @@ import { getCachedWords, setCachedWords } from '../../APIs/RedisAPI';
...
 import Version from '../../shared/constants/Version';
 import { SearchRegExp } from '../../shared/utils/createRegExp';
-import WordClassEnum from '../../shared/constants/WordClassEnum';
+import { Filters } from '../types';
+import { Keyword } from './types';
 
 type IgboSearch = {
   redisClient: RedisClientType | undefined;
-  keywords: { text: string; wordClass: WordClassEnum[]; regex: Pick<SearchRegExp, 'wordReg'> | SearchRegExp }[];
+  keywords: Partial<Keyword>[];
```
ijemmao, line 16: why are using `Partial` on `Keyword`? using `Partial` in this case will make other fields like `text` and `wordClass` optional, which is something we want to avoid doing since we want to require that when we're dealing with `Keyword` we can expect `text` and `wordClass` to exist - that's what the original type that you replaced was defining

src/controllers/utils/queries.ts
```diff
@@ -3,36 +3,48 @@ import { cjkRange } from '../../shared/constants/diacriticCodes';
 import WordClass from '../../shared/constants/WordClass';
 import Tenses from '../../shared/constants/Tenses';
 import StopWords from '../../shared/constants/StopWords';
+import { Keyword } from './types';
+import { SearchRegExp } from '../../shared/utils/createRegExp';
+import { Filters } from '../types';
 
-const generateMultipleNsibidi = (keywords) => keywords.map(({ text }) => ({ 'definitions.nsibidi': text }));
+type Keywords = Partial<Keyword>[];
```
ijemmao, line 10: again, we don't want to use `Partial` on `Keyword`

```suggestion
type Keywords = Keyword[];
```

ijemmao (changes requested):
just one last comment! really close~

src/controllers/utils/searchWordUsingIgbo.ts
zanio: The initial definition of keyword I had was the reason for this Partial usage. I have made update.

src/controllers/utils/queries.ts
```diff
@@ -75,15 +105,15 @@ const definitionsQuery = ({ regex, searchWord = '', filters }) => ({
...
   $or: [{ igbo: regex.wordReg }, { english: regex.definitionsReg }],
 });
 export const searchIgboTextSearch = fullTextSearchQuery;
 export const searchDefinitionsWithinIgboTextSearch = fullTextDefinitionsSearchQuery;
 /* Since the word field is not non-accented yet,
  * a strict regex search for words has to be used as a workaround */
-export const strictSearchIgboQuery = (keywords) => ({
-  $or: keywords.map(({ regex }) => ({ word: { $regex: regex.wordReg } })),
+export const strictSearchIgboQuery = (keywords: Keywords) => ({
+  $or: keywords.map(({ regex }) => ({ word: { $regex: regex?.wordReg } })),
```
ijemmao, line 116: nit: remove `?.` - it's not necessary and increases our build size

src/controllers/utils/queries.ts
```diff
@@ -3,36 +3,48 @@ import { cjkRange } from '../../shared/constants/diacriticCodes';
...
-  return { 'dialects.word': { $regex: regex.wordReg.source, $options: 'i' } };
+  return { 'dialects.word': { $regex: regex?.wordReg.source, $options: 'i' } };
 };
 
-const generateMultipleTensesWordRegex = (keywords) => {
+const generateMultipleTensesWordRegex = (keywords: Keywords) => {
   const tenses = Object.values(Tenses).map(({ value }) => {
     const { regex } = keywords[0];
-    return { [`tenses.${value}`]: { $regex: regex.wordReg.source, $options: 'i' } };
+    return { [`tenses.${value}`]: { $regex: regex?.wordReg.source, $options: 'i' } };
```
ijemmao, line 33: remove `?.`

src/controllers/utils/queries.ts
```diff
@@ -3,36 +3,48 @@ import { cjkRange } from '../../shared/constants/diacriticCodes';
...
   const { regex } = keywords[0];
-  return { variations: { $in: [regex.wordReg.source] } };
+  return { variations: { $in: [regex?.wordReg.source] } };
 };
 
-const generateMultipleDialectsWordRegex = (keywords) => {
+const generateMultipleDialectsWordRegex = (keywords: Keywords) => {
   const { regex } = keywords[0];
-  return { 'dialects.word': { $regex: regex.wordReg.source, $options: 'i' } };
+  return { 'dialects.word': { $regex: regex?.wordReg.source, $options: 'i' } };
```
ijemmao, line 27: remove `?.`

src/controllers/utils/queries.ts
```diff
@@ -3,36 +3,48 @@ import { cjkRange } from '../../shared/constants/diacriticCodes';
...
-  'definitions.definitions': { $in: compact(keywords.map(({ regex }) => regex.definitionsReg)) },
+const generateMultipleDefinitionsRegex = (keywords: Keywords) => ({
+  'definitions.definitions': { $in: compact(keywords.map(({ regex }) => regex?.definitionsReg)) },
 });
 
-const generateMultipleVariationsRegex = (keywords) => {
+const generateMultipleVariationsRegex = (keywords: Keywords) => {
   const { regex } = keywords[0];
-  return { variations: { $in: [regex.wordReg.source] } };
+  return { variations: { $in: [regex?.wordReg.source] } };
```
ijemmao, line 22: remove `?.`

src/controllers/utils/queries.ts
```diff
@@ -3,36 +3,48 @@ import { cjkRange } from '../../shared/constants/diacriticCodes';
...
+import { Filters } from '../types';
 
-const generateMultipleNsibidi = (keywords) => keywords.map(({ text }) => ({ 'definitions.nsibidi': text }));
+type Keywords = Keyword[];
+const generateMultipleNsibidi = (keywords: Keywords) => keywords.map(({ text }) => ({ 'definitions.nsibidi': text }));
 
-const generateMultipleWordRegex = (keywords) =>
-  keywords.map(({ regex }) => ({ word: { $regex: regex.wordReg.source, $options: 'i' } }));
+const generateMultipleWordRegex = (keywords: Keywords) =>
+  keywords.map(({ regex }) => ({ word: { $regex: regex?.wordReg.source, $options: 'i' } }));
```
ijemmao, line 14: remove `?.`

ijemmao (changes requested):
now that we're using `Keywords` we want to remove the unnecessary `?.` when accessing `wordReg`

ijemmao (approved):

PR #765, opened 2024-02-25
chore(design): remove Babel and update homepage with more modern styles

ijemmao:
## Background
Updating the Igbo API website to match the design system of the IgboSpeech website.

### Screenshots
image
image
image
image
image
image


ebubae:
looking at this late but this PR looks great! huge change but should be well worth it. Let me know if you're every interested in deploying with Vercel in the future!

Issue #781, opened 2024-04-24
Use code block for `style=no_style` code snippet in docs


ijemmao:
**Describe the bug**
The code snippet for an example parameter for the Example route isn't using a code block and looks like plain text.

**To Reproduce**
1. Go to https://igboapi.com/docs/endpoints/examples
2. See the example code snippet not using a code block

**Expected behavior**
The style=no_style should be in a code block like `style=no_style`

**Screenshots**
![image](https://github.com/nkowaokwu/igbo_api/assets/16169291/bd2bb45a-eb5e-4749-81b8-be64470e4956)

chidiadi01:
I'd love to work on this

ijemmao:
@chidiadi01 I assigned you to this ticket!

PR #783, opened 2024-05-20
Modified the style=no_style so it is in a code block

chidiadi01:
## Describe your changes

Made the *style=no_style* into a code block form _`style=no_style`_

This is to close issue #781 

## Something else to note
I also noticed that there were extra table columns in that row: standard, proverb, and biblical. They didn't render probably because they only exist on that row. It's what caused the code block problem in the first place. Maybe there's information that was supposed to be in there that was omitted? I'm creating a new issue based on that.


ijemmao:
Great find! Thank you~

PR #794, opened 2024-06-22
Updated API docs with information about proverbs

### LINKED ISSUE #782: Docs: Update the API docs with information about proverbs
## Add information about requesting proverbs from the Igbo API

[📑 Docs]: 

## What will this task cover?

- Adding information about proverbs to the API docs

chidiadi01:
## Describe your changes
Added a little guidance on how to request for proverbs using the Igbo API.

Closed  #782 

## Motivation and Context

The API is able to handle requests for proverbs, but it has not been explicitly written anywhere how to go about it. So, I believe developers may just think it's a feature in the works. It was necessary to add it.


src/pages/docs/endpoints/examples.mdx
```diff
@@ -79,6 +79,51 @@ A single example JSON object will be returned.
...
+ For example, this is a GET request for three (3) proverbs about `war`.
+
+ **Parameters**
+
+| Name   | Description | Example |
+| :----- | :----- | :----- |
+| **keyword** *required* | Keyword for querying words |  `keyword=war` |
+| **style** | Style for requested example sentences | `style=proverb` |
+| **page** | Page for results | `page=1` |
+| **range** | Page for result using [x,y] syntax, max range includes 25 documents | `range=[0, 2]` |
```
ijemmao, line 96: I feel like this section is a direct copy of the the "###### Get a Single Example" section but the only difference is that you provide the `style=proverb` field

Maybe we should consider adding a subsection in the "###### Get a Single Example" section that talks about proverbs?

chidiadi01: I just did that. I'll push so you can review it.

chidiadi01: I think that should be a better option. Though since my example had 3 examples, I placed it under the '###### Get Multiple Examples' section. 

I think the single example can be understood. Or do I need to add that too?

ijemmao: @chidiadi01 this is moving in the right direction! but we are still duplicating a lot of text.

there's a typo that needs to be fixed to make this PR mergeable.

screenshot of the typo: 
image

you can see that we list all valid values of what can be provided in the `style` parameter, but those values don't appear in the rendered markdown.

rather than adding the next lines you have included in this PR, you should include those extra parameter values inside the backticks (i.e. `style=no_stye|standard|proverb|biblical`)

src/pages/docs/endpoints/examples.mdx
```diff
@@ -53,6 +53,55 @@ Get examples from dictionary
 
  ***
 
+ **Get Poverbs**
```
ijemmao, line 56: remove all these lines and update the markdown typo in line 32

ijemmao (changes requested):

src/pages/docs/endpoints/examples.mdx
chidiadi01: Should I also remove the lines for the response of proverbs?

src/pages/docs/endpoints/examples.mdx
```diff
@@ -81,6 +129,7 @@ A single example JSON object will be returned.
 }
  ```
 
+
```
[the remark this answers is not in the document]
chidiadi01: Done

ijemmao (approved):

PR #812, opened 2024-10-09
Add translation endpoint to Igbo API

ebubae:
## Describe your changes
This PR updates the IgboAPI with a new endpoint to do Igbo-to-English text translation.

## Motivation and Context
We're constantly trying to improve our AI product offering to include more useful Igbo technology. This now includes a machine translation endpoint served by Nkọwa okwu team.

## How Has This Been Tested?
Tests can be found in `src/controllers/__tests__/translation.test.ts`


src/config.ts
```diff
@@ -26,6 +26,9 @@ const SENDGRID_NEW_DEVELOPER_ACCOUNT_TEMPLATE_SOURCE = defineString(
 // Igbo API
 const ENV_MAIN_KEY = defineString('ENV_MAIN_KEY').value();
 
+// Nkọwa Okwu AI Models
```
ijemmao, line 29: nit: it's Nkọwa okwu

src/config.ts
```diff
@@ -26,6 +26,9 @@ const SENDGRID_NEW_DEVELOPER_ACCOUNT_TEMPLATE_SOURCE = defineString(
...
+const ENV_IGBO_TO_ENGLISH_URL = defineString('ENV_IGBO_TO_ENGLISH_URL').value();
```
ijemmao, line 30: i'm mainly leaving this note to remind us to assign this environment variable in the cloud function envs list on GCP

src/routers/routerV2.ts
```diff
@@ -1,12 +1,23 @@
...
 import validId from '../middleware/validId';
 import validateApiKey from '../middleware/validateApiKey';
 import analytics from '../middleware/analytics';
 import attachRedisClient from '../middleware/attachRedisClient';
+import { MiddleWare } from 'src/types';
+
+const ONE_DAY = 24 * 60 * 60 * 1000;
+const REQUESTS_PER_MS_TRANSLATION = 5;
+
+const translationRateLimiter: MiddleWare = rateLimit({
```
ijemmao, line 17: i think we want rate limiting to prevent too many subsequent requests in period of time rather than limiting the total number of requests that can be made. the logic that's responsible for capping the total number of requests that the current user can make with the API Key lives in `authorizeDeveloperUsage`. 

i do think we want a rate limiter in general - we can talk about that more. 

also, can we have this rate limiter not be applied if the incoming request comes from our official platforms? we would want to check to see if the `X-API-Key` header is equal to `MAIN_KEY` which is in `config.ts`

src/routers/routerV2.ts
```diff
@@ -1,12 +1,23 @@
 import { Router } from 'express';
+import rateLimit from 'express-rate-limit';
 import { getWords, getWord } from '../controllers/words';
 import { getExample, getExamples } from '../controllers/examples';
 import { getNsibidiCharacter, getNsibidiCharacters } from '../controllers/nsibidi';
 import { getTranscription } from '../controllers/speechToText';
+import { getTranslation } from '..//controllers/translation';
```
ijemmao, line 7:
```suggestion
import { getTranslation } from '../controllers/translation';
```

just confirming if we want this as the path

src/routers/routerV2.ts
```diff
@@ -26,6 +37,13 @@ routerV2.get(
 
 // Speech-to-Text
 routerV2.post('/speech-to-text', analytics, validateApiKey, getTranscription);
+routerV2.post(
+  '/igbo-to-english',
```
ijemmao, line 41: i think we should name this route to `/translate` to keep it language agnostic for when we want to support translating between other native Nigerian languages

src/controllers/translation.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Talks to Igbo-to-English translation model to translate the provided text.
+ * @param req
+ * @param res
+ * @param next
+ * @returns English text translation of the provided Igbo text
+ */
+export const getTranslation: MiddleWare = async (req, res, next) => {
+  try {
+    const igboText: string = req.body['igbo'];
+    if (igboText.length == 0) {
```
ijemmao, line 23:
```suggestion
    if (Boolean(igboText)) {
```

with this check we don't run into the potential case where `igboText === undefined` and then it throws an error because we're indexing with `.length`

src/controllers/translation.ts
```diff
@@ -0,0 +1,50 @@
...
+      console.log('Cannot translate empty string');
```
ijemmao, line 24: we can remove these error related console statements since we're throwing them

src/controllers/translation.ts
```diff
@@ -0,0 +1,50 @@
...
+      throw new Error('Cannot translate empty string');
+    }
+
+    if (igboText.length > 120) {
```
ijemmao, line 28: can we assign this `120` to a constant value and then add a comment why it's set to 120?

src/controllers/translation.ts
```diff
@@ -0,0 +1,50 @@
...
+      method: 'POST',
+      url: `${IGBO_TO_ENGLISH_API}`,
+      headers: {
+        'Content-Type': 'application/json',
+        'X-API-Key': MAIN_KEY,
+      },
+      data: payload,
+    });
+
+    return res.send({ translation: response.translation });
```
ijemmao, line 46: super clean 🧼

src/routers/routerV2.ts
```diff
@@ -26,6 +37,13 @@ routerV2.get(
...
+  translationRateLimiter,
+  analytics,
+  validateApiKey,
```
ijemmao, line 44: we can talk about this more on our call, but you will want to update `authorizeDeveloperUsage` so that it will also count usage for the currently used API Key. 

You can look at `ApiUsageLimit.ts` which includes all the usage limits we allow for regular developer accounts.

ijemmao (commented):
this looks really goooooood! amazing first PR, you got like 90% off jump

src/controllers/translation.ts
```diff
@@ -0,0 +1,50 @@
...
+  } catch (err) {
+    return next();
```
ijemmao, line 48:
```suggestion
    return next(err);
```

express will run the error handler if we pass in `err` into `next`

src/config.ts
ijemmao: done ✅

src/routers/routerV2.ts
ebubae: After our discussion, I made this ticket for us to push a fix for rate limiting in a separate PR https://github.com/nkowaokwu/igbo_api/issues/814

src/routers/routerV2.ts
ebubae: honestly surprised this even worked

src/controllers/translation.ts
ebubae: updated to `if(igboText)` after discussion

src/controllers/translation.ts
ebubae: done!

src/config.ts
```diff
@@ -101,11 +104,12 @@ export const CORS_CONFIG = {
 export const API_ROUTE = isProduction ? '' : `http://localhost:${PORT}`;
 export const API_DOCS = 'https://docs.igboapi.com';
 
-// IgboSpeech
+// Nkọwa okwu AI Models
 export const SPEECH_TO_TEXT_API = isProduction
   ? 'https://speech.igboapi.com'
   : 'http://localhost:3333';
 
+export const IGBO_TO_ENGLISH_API = ENV_IGBO_TO_ENGLISH_URL;
```
ijemmao, line 112: nit: can you move this line on line up to line 111? That way there's an empty new line before line 113

src/routers/routerV2.ts
```diff
@@ -1,12 +1,24 @@
...
+import { MiddleWare } from 'src/types';
+
+// TODO: add rate limiting with upstash
+// const ONE_DAY = 24 * 60 * 60 * 1000;
+// const REQUESTS_PER_MS_TRANSLATION = 5;
+
+// const translationRateLimiter: MiddleWare = rateLimit({
+//   windowMs: ONE_DAY,
+//   max: REQUESTS_PER_MS_TRANSLATION,
+// });
```
ijemmao, line 21: you can delete this comment since you have a corresponding GitHub issue

src/middleware/helpers/authorizeDeveloperUsage.ts
```diff
@@ -58,8 +59,27 @@ export const authorizeDeveloperUsage = async ({
...
+}) => {
+  const extractedPath = getPath(path);
+  return handleDeveloperUsage({
     developer,
-    apiType: path.startsWith('speech-to-text') ? ApiType.SPEECH_TO_TEXT : ApiType.DICTIONARY,
+    apiType: getApiTypeFromRoute(extractedPath),
   });
+};
+
+const getApiTypeFromRoute = (route: string): ApiType => {
```
ijemmao, line 70: nit: can you add a comment for this function?

src/controllers/__tests__/translation.test.ts
```diff
@@ -0,0 +1,62 @@
...
+describe('translation', () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+  it('calls Nkọwa okwu ML translation server', async () => {
+    const req = requestFixture({
+      body: { igbo: 'aka' },
+      headers: {
+        'Content-Type': 'application/json',
+        'X-API-Key': 'main_key',
```
ijemmao, line 18: i think we should import `MAIN_KEY` instead of hardcoding `'main_key'`

src/controllers/__tests__/translation.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      },
+    });
+    const res = responseFixture();
+    const next = nextFunctionFixture();
+    jest.spyOn(axios, 'request').mockResolvedValue({
+      data: { igbo: 'aka'.repeat(100) },
+    });
+    await getTranslation(req, res, next);
+    expect(next).toHaveBeenCalled();
```
ijemmao, line 44: is it possible to check the error message here too so we know which error was thrown?

src/controllers/__tests__/translation.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        'X-API-Key': 'main_key',
+      },
+    });
+    const res = responseFixture();
+    const next = nextFunctionFixture();
+    jest.spyOn(axios, 'request').mockResolvedValue({
+      data: { igbo: '' },
+    });
+    await getTranslation(req, res, next);
+    expect(next).toHaveBeenCalled();
```
ijemmao, line 60: same thing here

ijemmao (approved):
LGTM!!! Just a few nit comments about spacing, using variables, and adding comments - really excited for this 🚀

src/routers/routerV2.ts
ebubae: done!

src/controllers/__tests__/translation.test.ts
ebubae: now doing that check!

src/controllers/translation.ts
```diff
@@ -0,0 +1,52 @@
+import axios from 'axios';
+import { MiddleWare } from '../types';
+import { IGBO_TO_ENGLISH_API, MAIN_KEY } from '../config';
+
+interface TranslationMetadata {
+  igbo: string;
```
ijemmao, line 6: I think we would want to have a schema similar to something like the following:
``` suggestion
  text: string;
  language: LanguageEnum
```

`LanguageEnum` would include he ISO language codes, so for Igbo it would be `ibo`

ebubae:
I've added the schema changes we discussed. Also some new validation for the schema values as well as new tests for the extra validation. Let me know if any more needs to be done!

src/controllers/translation.ts
ebubae: I'm now validating the request body. For now, the call to the ML endpoint won't change but that's more of an implementation detail.

ijemmao (approved):
excited!!! this is going to make for a great demo

PR #822, opened 2024-11-11
English to Igbo Translation Endpoint

ebubae:
## Describe your changes
Add English to Igbo Translation to the `/translate` endpoint.

## Motivation and Context
We've recently added Igbo to English translation in #812. This PR updates the code to support English to Igbo translation as well.


src/controllers/translation.ts
```diff
@@ -32,43 +37,48 @@ const IGBO_ENGLISH_TRANSLATION_INPUT_MAX_LENGTH = 120;
...
-    const igboText = requestBody.text;
-    if (!igboText) {
+    const textToTranslate = requestBody.text;
+    if (!textToTranslate) {
       throw new Error('Cannot translate empty string');
     }
-
-    if (igboText.length > IGBO_ENGLISH_TRANSLATION_INPUT_MAX_LENGTH) {
-      throw new Error('Cannot translate text greater than 120 characters');
+    const translationKey =
```
ijemmao, line 65: Instead of casting these strings together to create a key to determine the payload shape, I think the `sourceLanguageCode` could map to the desired payload shape. Something like

```ts
const PayloadKeyMap = {
  [LanguageEnum.Igbo]: 'igbo',
  [LanguageEnum.English]: 'english'
}
```

That way, we could take the value of indexing into that map to determine the final payload shape:

```ts
const payload = { [PayloadKeyMap[requestBody.sourceLanguageCode]: textToTranslate }
```

If `requestBody.sourceLanguageCode === 'ibo'` then we would get a final payload of `{ igbo: textToTranslate }` (if my logic is correct 😅 )

I think there's an even cleaner way of us handling this, but I think relying on maps rather than strings to determine payload shapes is a more stable way to go

src/controllers/translation.ts
```diff
@@ -19,9 +15,18 @@ export interface Translation {
   translation: string;
 }
 
-// Due to limit on inputs used to train the model, the maximum
-// Igbo translation input is 120 characters
-const IGBO_ENGLISH_TRANSLATION_INPUT_MAX_LENGTH = 120;
+// TODO: move this information to remote config
+const SUPPORTED_TRANSLATIONS = {
```
ijemmao, line 19: I think we would want to update this map to rely on nested maps for better indexing:

```ts
const SUPPORTED_TRANSLATIONS = {
  [LanguageEnum.IGBO]: {
    [LanguageEnum.ENGLISH]: {
      maxInputLength: 120,
      translationAPI: IGBO_TO_ENGLISH_API,
    }
  },
  [LanguageEnum.ENGLISH]: {
    [LanguageEnum.IGBO]: {
      maxInputLength: 120,
      translationAPI: ENGLISH_TO_IGBO_API,
    }
  }
}
```

So then when we use `SUPPORT_TRANSLATIONS` we can index using our `LanguageEnum` values:

```ts
const translation = SUPPORTED_TRANSLATIONS[requestBody.sourceLanguageCode][requestBody.destinationLanguageCode] // === { maxInputLength: 120, translationAPI: IGBO_TO_ENGLISH_API }
```

ijemmao: I'm also interested to hear why we would want to move this to RemoteConfig! I think it's fine having this object here, but I might be missing something

ijemmao (commented):
Looking really good! Just a few comments about strings vs. maps

src/controllers/translation.ts
ebubae: Switched to something like this. Thanks for the suggestion!

src/controllers/translation.ts
ebubae: I ended up getting something to work that typescript could actually be happy with by indexing all the languages on the first level of the object. Inner languages don't need to map out each one as not to be overly exhaustive.

PR #823, opened 2024-12-03
Introduce Paywalls and add ownerFirebaseId on Projects

ijemmao:
## Background
We have a new collection called `Paywall` that tracks the limits applied to certain users on [Sabbi](https://dashboard.sabbidata.com). A migration script has been created to create `Paywall` documents for each project currently existing in Sabbi. Additionally, we want to track the owning user on the `Project` collection, so another migration script is created to add `ownerFirebaseId` on the `Project` collection.


ebubae (approved):
migration looks good to me

PR #832, opened 2025-06-13
chore: add sabbidata to footer links

Ebugo:
## Describe your changes
Added sabbidata to the footer links

## Motivation and Context
We needed to point to Sabbi within the footer on https://igboapi.com/, just below Nkọwa okwu and Chrome Extension

## How Has This Been Tested?
Updated the unit test 'Footer' to accommodate the text added "Sabbi"

## Screenshots
image

image

### DIFF
--- a/src/pages/components/Footer/Footer.tsx
+++ b/src/pages/components/Footer/Footer.tsx
@@ -23,6 +23,7 @@ import {
   LINKEDIN,
   NKOWAOKWU,
   NKOWAOKWU_CHROME,
+  SABBI_HOME,
   TWITTER,
   YOUTUBE,
 } from '../../../siteConstants';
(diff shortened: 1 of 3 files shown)


ijemmao:
@Ebugo great work! can you add a screenshot of your change for both desktop and mobile view to confirm this looks good?

Ebugo:
> @Ebugo great work! can you add a screenshot of your change for both desktop and mobile view to confirm this looks good?

Done

tohuw:
Investigated the failing **Build and Test (20.x, 6)** check (the run's logs have since expired, so I reproduced locally):

- Your change is not the cause. Checked out this branch on current master: the `Footer` suite **passes** with your added assertion, and the full frontend suite fails on exactly the same 4 suites (`Demo`, `Navbar`, `NavigationOptions`, `App`) with or without your diff — it's dependency drift on master (`@chakra-ui/icons` needs `createIcon` from the manual Chakra mock, and `@zag-js/focus-visible` crashes under newer jsdom).
- Fix for the underlying breakage is up in #835. Once that merges, a rebase/re-run here should go green.

---
*This comment was created by Claude Fable 5 on behalf of [@tohuw](https://github.com/tohuw).*

